PR #1, opened 2016-10-05
Normalise registration model

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-122)


registration/queue.py
```diff
@@ -87,10 +87,11 @@ def is_valid_registration(message_body):
         """
         try:
             registration = json.loads(message_body)
-        except (ValueError, json.decoder.JSONDecodeError):
```
richtier: `json.decoder.JSONDecodeError` is sometimes in `json.JSONDecodeError`. Fortunately `JSONDecodeError` is a subclass of `ValueError`, so `ValueError` will catch both.

registration/queue.py
```diff
@@ -144,11 +144,12 @@ def save_registration(self, sqs_message_id, registration):
         logger.debug(
             "Saving new registration from message '{}'".format(sqs_message_id)
         )
+        data = registration.copy()
```
richtier: copying `registration` to prevent mutations to it affecting things outside of this function. Avoiding side-effects.

registration/queue.py
```diff
@@ -144,11 +144,12 @@ def save_registration(self, sqs_message_id, registration):
...
+        data['sqs_message_id'] = sqs_message_id
+        serializer = serializers.RegistrationSerializer(data=data)
```
richtier: instead of manually calling `models.Registration.objects.create(` and unpacking the contents of `data` - lets use a serializer (below).

richtier:
@konradko @monikasulik happy with this?

registration/queue.py
monikasulik: I suggest making this a comment in the actual code :)

registration/queue.py
monikasulik: Again, this feels like a comment that belongs in the code :)

registration/queue.py
```diff
@@ -144,11 +144,12 @@ def save_registration(self, sqs_message_id, registration):
...
+        assert serializer.is_valid()
```
monikasulik: So what happens if this throws an exception (unlikely I guess because we're already validating data in ui on the form, but not entirely impossible)? I guess the backend will then fail to register the person? Will the user be informed of the issue or is this just something that will be logged on sentry? And most importantly - is what happens definitely the behaviour we want? :)

registration/queue.py
konradko: Why do we need ``registration.copy()``?

registration/models.py
```diff
@@ -3,20 +3,21 @@
 
 
 class Registration(models.Model):
-
+    aims = JSONField(help_text='list of strings e.g., ["AIM1", "AIM2"]')
+    company_number = models.CharField(max_length=8)
```
konradko: Help text saying "Companies House ID" would be useful

registration/models.py
```diff
@@ -3,20 +3,21 @@
...
     created = models.DateTimeField(auto_now_add=True)
-    data = JSONField()
+    email = models.EmailField()
+    personal_name = models.CharField(max_length=100)
```
konradko: Better to set max_length to 255 - no reason to limit to 100

registration/queue.py
konradko: this method is no longer needed - serializer on the view validates it all

registration/queue.py
konradko: What are the side-effects?

registration/queue.py
konradko: this is already called on the view - does not seem to be necessary to repeat it here.

monikasulik: I'd argue that data integrity is very important, so if it's not costly, another check is reasonable. And this is already checked in the view then it's probably fine to have this exception uncaught here.

konradko: this is 3rd time it gets validated - I'd argue 2 times is enough :)

registration/queue.py
richtier: the side effect would be `registration` would be updated is all scopes - not just in the current function.
```
def api_caller(data):
    data['version'] = 1
    return client.some_action(data)

details = {'name': 'jim'}
api_caller(details)
print(details) # shows {'name': 'jim', 'version': 1}
```

registration/queue.py
richtier: https://github.com/uktrade/directory-api/pull/1#discussion_r83023343 :)

registration/queue.py
richtier: we need to call is_valid to get the data from the serializer.

is_value should not return false because, as konrad said, validation happens in other places before this.

richtier:
@konradko @monikasulik I have made the required changes :)

monikasulik:
Hmm, turns out we both need to approve now for github to be happy?

PR #3, opened 2016-10-12
SSO prototype

konradko:
https://uktrade.atlassian.net/browse/ED-149


requirements.txt
```diff
@@ -1,7 +1,12 @@
 dj-database-url==0.4.1
-Django==1.10.2
+Django==1.9.10
```
monikasulik: How come you're downgrading this?

monikasulik:
Tests are failing (see make_docker step on CircleCI).

alice/middleware.py
```diff
@@ -13,24 +11,6 @@ def __init__(self):
         self.logger = logging.getLogger(__name__)
 
     def process_request(self, request):
-        if not self._test_signature(request):
+        if not SignatureRejection.test_signature(request):
```
richtier: nice composition :+1:

api/settings.py
```diff
@@ -20,35 +20,61 @@
...
+    'rest_framework_swagger',
+    'rest_framework.authtoken',
+    'rest_auth',
+    'rest_auth.registration',
+    'allauth',
+    'allauth.account',
+    'oauth2_provider',
+    'corsheaders',
+    'enrollment.apps.EnrollmentConfig',
+]
```
richtier: there are a lot more installed apps now. Is this because the auth stuff requires these apps?

api/settings.py
```diff
@@ -20,35 +20,61 @@
...
     'django.middleware.security.SecurityMiddleware',
-    'alice.middleware.SignatureRejectionMiddleware',
+    'django.contrib.sessions.middleware.SessionMiddleware',
     'django.middleware.common.CommonMiddleware',
+    'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
+    'corsheaders.middleware.CorsMiddleware',
 ]
 
+CORS_ORIGIN_ALLOW_ALL = True
```
richtier: why not use a whitelist? This feels like a security concern.

api/urls.py
```diff
@@ -1,14 +1,145 @@
...
+        r'^registration/$',
+        rest_auth.registration.views.RegisterView.as_view(),
+        name='rest_register'
+    ),
+
+    # TODO: override this view to handle it and send post to /verify-email/
+    # endpoint with proper key (see allauth.account.views.ConfirmEmailView for
+    # example)
+    url(
+        r'^account-confirm-email/(?P<key>[-:\w]+)/$',
```
richtier: Looks like we expect the token to contain the characters `[A-Za-z0-9_]` or `:` or `-`.

If the token contains an unexpected character (such as a space) then the request will 404.

If there is a risk of an unexpected characters occurring then a 400 response would be more helpful, so that would mean passing the token as a querystring param or in the body of the request.

If there is no risk of it happening then no change required.

enrollment/management/commands/csv.py
```diff
@@ -0,0 +1,36 @@
...
+
+from django.core.management.base import BaseCommand
+
+from enrollment import models
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        live_date = datetime.date(year=2016, month=7, day=21)
```
richtier: is this purposefully hard-coded?

enrollment/management/commands/csv.py
```diff
@@ -0,0 +1,36 @@
...
+
+        enrollment_dicts = []
+        keys = set()
+
+        for enrollment in enrollments:
+            enrollment_dict = {'created': enrollment.created}
+            enrollment_dict = json.loads(enrollment.data)
+            enrollment_dict.update(enrollment)
+            enrollment_dicts.append(enrollment_dict)
+            keys.update(set(enrollment.keys()))
```
richtier: `keys.update(enrollment.keys())` would work fine.

api/settings.py
konradko: yup

api/settings.py
konradko: yeah this is for test purposes only - will add a todo / ticket

api/urls.py
konradko: token is generated by the allauth, which consumes this url too - so no risk

enrollment/management/commands/csv.py
konradko: ye, will most likely remove this mgm command

konradko:
@monikasulik latest tests did pass: https://circleci.com/gh/uktrade/directory-api/35

PR #4, opened 2016-10-13
Company profile - model & serializer

monikasulik:
Adds:
* company app
* company model
* company model migration
* company serializer


company/models.py
```diff
@@ -0,0 +1,15 @@
...
+from django.contrib.postgres.fields import JSONField
+
+
+class Company(models.Model):
+
+    name = models.CharField(max_length=255)
+    number = models.CharField(max_length=8, validators=[MinValueValidator(8)])
+    website = models.URLField(max_length=255)
+    description = models.TextField()
+    aims = JSONField()
```
richtier: note `aims` and `number` are already represented on the [Entrolment model](https://github.com/uktrade/directory-api/blob/master/enrolment/models.py#L5). Perhaps remove them from the Enrolment model?

richtier: I think all these fields should be nullable except number. @konradko ?

company/serializers.py
```diff
@@ -0,0 +1,12 @@
...
+from company import models
+
+
+class CompanySerializer(serializers.ModelSerializer):
+
+    id = serializers.CharField(read_only=True)
+
+    class Meta(object):
+        model = models.Company
+        fields = ('name', 'number', 'website', 'description', 'aims')
```
richtier: please add a test for the serializer.

company/models.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __str__(self):
+        return self.name
```
richtier: please add a test for the model (e.g., the `__str__` method does its job)

company/models.py
richtier: Can you also add the logo field to the Company please.

monikasulik: Re: logo field, I was actually told not to add it as part of ED-145? Adding it does require sorting pillow (there may be non-python dependencies such as jpeg), storage and stuff.

company/models.py
monikasulik: I can certainly test the string method, but it doesn't really do anything functionality-related. This is really just a helper for devs when we experiment with stuff in the command line, use pdb etc.
The model itself is mainly syntax - this basically gets tested by flake8 and by attempting to run any kind of tests (if the model had incorrect syntax, django as a whole would fail).

company/serializers.py
monikasulik: As I'm basically just using DRF stuff and haven't written any logic of my own, surely the tests in the DRF library would cover anything that should be tested, no? That's generally been the policy everywhere I've worked so far.

company/models.py
richtier: @monikasulik re:logo field lets just use a filefield. We're not doing any image processing so dont need `ImageField`. Feel free to do is as part of 145 :+1:

company/models.py
richtier: I'm a fan of confirming that the model can be instantiated with errors. I've seen in the past instantiation fail because of __repr__ and __str__

If you think its overkill then feel free to leave it out :+1:

company/serializers.py
richtier: the serializer specified fields by their names `('name', 'number', ...)`. If we specifid a field that does not exist on the model the serializer will throw an exception.

For example if we did this: `('name', 'number', 'website' 'description', 'aims')` for example, it would fail because  'websitedescription' is not a field. I would prefer to catch that in tests than in prod :smile:

company/models.py
monikasulik: That is a very good point - I have indeed once seen a failure because ```__unicode__``` had some weird issue. Rare, but possible.

company/serializers.py
monikasulik: Fair enough, I'll write something super simple to test it.

monikasulik:
@richtier I'm struggling with the field changes you requested to enrolment. To make those changes I need to change the enrolment serializer as well (it needs to save to two models instead of one) and those changes don't seem quite so straightforward and require quite a few tests.
With this being the end of the sprint, I'm kind of running out of time for this. Are you ok with me creating a separate ticket to remove the aims and company number from enrolment and introduce a foreign key to company? We could include it in the next sprint.

company/models.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+class Company(models.Model):
+
+    name = models.CharField(max_length=255, blank=True)
+    number = models.CharField(max_length=8, validators=[MinLengthValidator(8)])
+    website = models.URLField(max_length=255, blank=True)
+    description = models.TextField(blank=True)
+    aims = JSONField()
+    logo = models.FileField(blank=True, null=True)
```
richtier: when the user completes registration they will only provide the `number` and `aims`. All these other fields need to be either `null=True`, or some special handling is required to turn the empty values into empty strings.

An implication of this is we will also need some more serializer tests confirming that the serializer works when only `number` and `aims` are provided,

monikasulik: Does commit 275200c1cac2b5cb50a1c7b10a86b380202915f9 not cover this? `blank=True` implies `default=''`, but I can set it explicitly on the serializer if you're more comfortable with that?

richtier: `blank=True` is for application level validation of the field. If we pass in `None`, the application will validate the value as OK, but when we try to save the value in the db then the database will throw non null constraint error.

konradko:
@richtier @monikasulik can we merge this?

monikasulik:
Nope, we can't. Discussed with Richard and there's a test I actually need to write that came out of our discussion.

PR #6, opened 2016-10-14
Company profile API views

monikasulik:
https://uktrade.atlassian.net/browse/ED-146

### DIFF
--- a/company/tests/test_views.py
+++ b/company/tests/test_views.py
@@ -1,0 +1,40 @@
+import json
+
+import pytest
+
+from django.core.urlresolvers import reverse
+
+from rest_framework.test import APIClient
+from rest_framework import status
+
+from company.models import Company
+from company.tests import VALID_REQUEST_DATA
+
+
+@pytest.mark.django_db
+def test_company_retrieve_view():
+    client = APIClient()
+    company = Company.objects.create(**VALID_REQUEST_DATA)
+
+    response = client.get(reverse('company', kwargs={'pk': company.pk}))
+
+    expected = {'id': str(company.id)}
+    expected.update(VALID_REQUEST_DATA)
+    assert response.status_code == status.HTTP_200_OK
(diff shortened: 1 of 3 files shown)


monikasulik:
@konradko This is now ready for another look.

company/tests/test_views.py
```diff
@@ -0,0 +1,56 @@
...
+from rest_framework import status
+
+from company.models import Company
+from company.tests import VALID_REQUEST_DATA
+
+
+@pytest.mark.django_db
+def test_company_retrieve_view():
+    client = APIClient()
+    company = Company.objects.create(**VALID_REQUEST_DATA)
```
konradko: Seeing how often ``VALID_REQUEST_DATA`` is used, maybe we should consider making a pytest 'fixture' out of them. I don't like pytests fixtures though, they are useful, but it's an odd pattern

company/views.py
```diff
@@ -0,0 +1,13 @@
...
+from company.serializers import CompanySerializer
+from company.models import Company
+
+
+class CompanyRetrieveUpdateAPIView(RetrieveUpdateAPIView):
+
+    model = Company
+    serializer_class = CompanySerializer
+
+    def get_queryset(self):
```
konradko: is this necessary? Just specifying the model should achieve that I believe

monikasulik: I was surprised, but I was getting exceptions when this wasn't here.

company/tests/test_views.py
monikasulik: I kind of copied what was in the codebase already in enrolment. If we're changing it then we need to change both places :)

konradko: yup, that's what pytest fixtures are for. Fine to leave it, like I mentioned, not a huge fan

company/views.py
konradko: that is surprising

PR #8, opened 2016-10-17
User Model & Serializer

monikasulik:
https://uktrade.atlassian.net/browse/ED-185

I still need to do another PR with an admin.py for both the user and company.


api/settings.py
```diff
@@ -188,6 +189,9 @@
 
 # Auth
 
+AUTH_USER_MODEL = 'user.User'
```
richtier: Thoughts on using a custom user model?

Its much much easier to use custom user model at start of project compared with when project is mature.

We would even define a custom user model with no extra logic - it will make us more flexible in future.

richtier:
I'll leave this to @konradko to review because there may be SSO implications I am unaware of.

user/models.py
```diff
@@ -0,0 +1,52 @@
...
+from django.contrib.auth.models import AbstractBaseUser
+from django.utils import timezone
+from django.utils.translation import ugettext_lazy as _
+
+
+class User(AbstractBaseUser):
+
+    # NOTE: AbstractBaseUser also implements two more fields:
+    # password and last_login. Because of our architecture
+    # last_login won't necessarily be reliable.
```
konradko: why? Django authentication will still be used, just with oauth2 integrated

user/models.py
```diff
@@ -0,0 +1,52 @@
...
+    name = models.CharField(
```
konradko: so not first / last after all?

konradko:
@monikasulik rebase and good to merge

user/models.py
konradko: the login / session stuff will still happen on api, it's just that UI will have it's own session with api-session-id stored in it :-)

PR #9, opened 2016-10-18
Split Enrolment creation into Enrolment, Company, and User

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-173)

I have added placed the Enrolment, the User, and the Company creation within one transaction so either all three are created or none are created - to avoid orphaned objects.

### DIFF
--- a/enrolment/queue.py
+++ b/enrolment/queue.py
@@ -2,14 +2,16 @@ import gc
 import json
 import logging
 
-from django.conf import settings
-from django.db import IntegrityError
-
 from psycopg2.errorcodes import UNIQUE_VIOLATION
+from rest_framework.serializers import ValidationError
+
+from django.conf import settings
+from django.db import IntegrityError, transaction
 
 from enrolment import serializers
 from enrolment.utils import ExitSignalReceiver, QueueService
-
+from user.serializers import UserSerializer
+from company.serializers import CompanySerializer
 
 logger = logging.getLogger(__name__)
 
@@ -104,9 +106,9 @@ class Worker:
             "Processing message '{}'".format(message.message_id)
         )
         if self.is_valid_enrolment(message.body):
-            self.save_enrolment(
-                sqs_message_id=message.message_id,
-                enrolment=json.loads(message.body)
+            self.create_objects(
+                json_payload=message.body,
+                message_id=message.message_id,
             )
         else:
             logger.error(
@@ -134,21 +136,55 @@ class Worker:
             exception.pgcode == UNIQUE_VIOLATION
         )
 
-    def save_enrolment(self, sqs_message_id, enrolment):
+    @transaction.atomic
+    def create_objects(self, json_payload, message_id):
+        # This function will create either all three objects (Enrolment, User,
+        # Company), or none of them. Using transactions.atomic rollsback our db
+        # changes on uncaught exception - so if `save_company` raises an
+        # exception, User and Enrolment objects will not be created too.
+        payload = json.loads(json_payload)
+        self.save_enrolment(
+            sqs_message_id=message_id,
+            aims=payload['aims'],
+            company_number=payload['company_number'],
+            email=payload['email'],
+            personal_name=payload['personal_name'],
+        )
+        self.save_user(
+            email=payload['email'],
+            name=payload['personal_name'],
+            referrer=payload['referrer'],
+            password=payload['password'],
+        )
+        # TODO: ED-228
+        # Link the company to the user.
+        self.save_company(
+            aims=payload['aims'],
+            number=payload['company_number'],
+        )
+
+    def save_enrolment(
+        self, sqs_message_id, aims, company_number, email, personal_name
+    ):
         """Creates new enrolment.models.Enrolment
 
         Args:
             sqs_message_id (str): SQS message ID
-            enrolment (str): Enrolment
+            aims (str[]): Goals of joining the scheme
+            company_number (str): Companies House number
+            email (str): User's email
+            personal_name (str): User's full name
         """
         logger.debug(
             "Saving new enrolment from message '{}'".format(sqs_message_id)
         )
-        # `copy` enrolment to avoid it being be mutated is other scopes.
-        data = enrolment.copy()
-        data['sqs_message_id'] = sqs_message_id
-
-        serializer = serializers.EnrolmentSerializer(data=data)
+        serializer = serializers.EnrolmentSerializer(data={
+            'aims': aims,
+            'company_number': company_number,
+            'email': email,
+            'personal_name': personal_name,
+            'sqs_message_id': sqs_message_id,
+        })
         assert serializer.is_valid()
 
         try:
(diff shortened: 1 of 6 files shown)


enrolment/queue.py
```diff
@@ -134,21 +134,51 @@ def is_postgres_unique_violation_error(exception):
             exception.pgcode == UNIQUE_VIOLATION
         )
 
-    def save_enrolment(self, sqs_message_id, enrolment):
+    @transactions.atomic
```
konradko: I understand this prevents the creation of user and company on ``IntegrityError`` from ``save_enrolment``. Wonder if we could make it a bit more explicit, as it's not immediately obvious. Good approach though!

richtier: it prevents creation on any Exception (I think), which is super helpful. Will add a comment (and confirm my thoughts).

monikasulik:
I like the approach as well - nice and clean. One question though - we will be removing the enrolment model completely, right?

richtier:
@monikasulik not for alpha. One useful feature it provides us is it enforces uniqueness of the amazon SQS message id, preventing the same message being processed twice.

Eventually we will remove it though.

enrolment/queue.py
```diff
@@ -162,3 +199,37 @@ def save_enrolment(self, sqs_message_id, enrolment):
                 )
             else:
                 raise
+
+    def save_user(self, email, name, referrer, plaintext_password):
+        serializer = UserSerializer(data={
+            'email':email,
+            'name': name,
+            'referrer': referrer,
+            'password': make_password(plaintext_password),
```
richtier: @konradko currently the `directory-api` does not hash the password before it sends the password to `directory-api`. That has the advantage of allowing authentication with clients other than `directory-ui`

enrolment/tests/test_queue.py
```diff
@@ -32,3 +35,95 @@ def test_process_message(self):
...
+                mock.Mock(message_id='1', body=VALID_REQUEST_DATA_JSON)
+            )
+        assert Company.objects.count() == 0
+        assert User.objects.count() == 0
+        assert enrolment.models.Enrolment.objects.count() == 0
+
+    @pytest.mark.django_db
+    @mock.patch.object(enrolment.queue.Worker, 'save_enrolment',
+                       mock.Mock(side_effect=Exception))
+    def test_process_message_enrolment_exception_rollback(self):
```
monikasulik: As an extra precaution, I'd quite like a set of these tests that mock ValidationError for serializer.is_valid()
The reason for this is that we're catching those errors and then reraising them. Want to make sure that nobody accidentally deletes the raising line at some point :)

enrolment/tests/test_queue_worker_command.py
```diff
@@ -10,7 +11,14 @@
...
 
 import enrolment.queue
-from enrolment.tests import MockBoto, VALID_REQUEST_DATA_JSON
+from enrolment.tests import MockBoto, VALID_REQUEST_DATA
+
+
+def get_reqest_data_json(email_preix):
+    request_data = VALID_REQUEST_DATA.copy()
+    # email muse be unique.
+    request_data['email'] = email_preix + request_data['email']
```
monikasulik: typo: email_prefix rather than preix

enrolment/tests/test_queue_worker_command.py
```diff
@@ -20,7 +28,7 @@ def test_terminates_gracefully_while_processing_messages(self):
         """ Test queue worker stops running on sigterm """
         worker = enrolment.queue.Worker()
         worker.enrolment_queue._queue.receive_messages.return_value = [
-            mock.Mock(message_id=x, body=VALID_REQUEST_DATA_JSON)
+            mock.Mock(message_id=x, body=get_reqest_data_json(str(x)))
```
monikasulik: typo: request rather than reqest

user/serializers.py
```diff
@@ -10,7 +10,8 @@ class UserSerializer(serializers.ModelSerializer):
     class Meta(object):
         model = models.User
         fields = ('id', 'name', 'email', 'terms_agreed', 'referrer',
-                  'date_joined')
+                  'date_joined', 'password',)
+        extra_kwargs = {'password': {'write_only': True, 'required': False}}
```
monikasulik: We probably should have a test for these in user/test/test_serializers.py

enrolment/queue.py
```diff
@@ -2,14 +2,17 @@
...
 
-from django.conf import settings
-from django.db import IntegrityError
-
 from psycopg2.errorcodes import UNIQUE_VIOLATION
+from rest_framework.serializers import ValidationError
+
+from django.conf import settings
+from django.db import IntegrityError, transaction
+from django.contrib.auth.hashers import make_password
```
richtier: good question. FWIW, for background, when we call `User.set_password`, django calls [`make_password`](https://github.com/django/django/blob/master/django/contrib/auth/base_user.py#L105).

enrolment/queue.py
```diff
@@ -134,31 +137,117 @@ def is_postgres_unique_violation_error(exception):
...
+            'company_number': company_number,
+            'email': email,
+            'personal_name': personal_name,
+            'sqs_message_id': sqs_message_id,
+        })
         try:
+            serializer.is_valid()
             serializer.save()
+        except ValidationError:
+            logger.error(
```
monikasulik: Ideally `logger.exception` in this case - exception additionally logs the whole traceback also.

konradko:
@richtier @monikasulik I don't see us removing the Enrolment model tbh, unless we would stop using SQS, which we do not expect

enrolment/tests/test_queue.py
```diff
@@ -32,3 +35,97 @@ def test_process_message(self):
...
+                    worker.process_message(
+                        mock.Mock(message_id='1', body=VALID_REQUEST_DATA_JSON)
+                    )
+                assert Company.objects.count() == 0
+                assert User.objects.count() == 0
+                assert enrolment.models.Enrolment.objects.count() == 0
+
+    @pytest.mark.django_db
+    def test_process_message_user_exception_rollback(self):
+        for exception_class in [Exception, ValidationError]:
```
monikasulik: I'm fairly sure this still wouldn't catch deleting the raise line (a ValidationError is a subclass of Exception so this actually shouldn't matter), but you can try it and see what happens :) The issue is what you're mocking - i.e. for save_user you'd need to do an additional test where you mock user.serializers.UserSerializer.save or is_valid and see what happens when that particular method throws ValidationError.

user/serializers.py
monikasulik: Did these make it in? (I was thinking a test that checks if password is indeed optional and perhaps a test which checks that password cannot be read or whatever the write_only flag does exactly)

enrolment/tests/test_queue.py
monikasulik: Just in case I haven't explained well, what I mean is that if for example this line were accidentally deleted: https://github.com/uktrade/directory-api/pull/9/files#diff-d5915164fc18b4024e221bec0778fef1R234 then I don't think our tests would catch that we're now happily getting enrolment and company saved without the user.

enrolment/queue.py
richtier: how would the traceback would add value here> We're logging the serializer errors. I think that is enough.

enrolment/tests/test_queue.py
monikasulik: ok, fair enough :)

user/serializers.py
richtier: yep, they are in now :)

enrolment/queue.py
monikasulik: I don't think it's particularly important either way. I would tend to just do `exception` rather than `error` in `except` clauses simply because I think it's better to have too much info than too little when debugging (one never quite knows what and how things might go wrong). But yeah, it's fairly unlikely that a full traceback would give us any info that would be crucial to debug this. serializer.errors is the most valuable.

PR #10, opened 2016-10-18
Fix docker test and requirements

konradko:


richtier:
looks ok to me, but I'm not a docker person so YMMV

PR #11, opened 2016-10-18
Link company to user

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-228)

Once `ED-173-split-enrolment` is merged I shall re-open this against `master`.


company/models.py
```diff
@@ -17,6 +18,7 @@ class Company(models.Model):
     description = models.TextField(blank=True, null=True, default='')
     aims = JSONField()
     logo = models.FileField(blank=True, null=True)
+    user = models.ForeignKey(User)
```
monikasulik: Is it possible to have this blank or will every company in our db definitely always have a user (if the only way a company finds itself in our db is through registration then this is fine, but I think I heard something about somekind of migration of data from a previous system?)

company/serializers.py
```diff
@@ -9,7 +9,15 @@ class CompanySerializer(serializers.ModelSerializer):
...
     class Meta(object):
         model = models.Company
-        fields = ('id', 'name', 'number', 'website', 'description', 'aims')
+        fields = (
+            'aims',
+            'description',
+            'id',
+            'name',
+            'number',
+            'user',
```
monikasulik: The company serializer tests need to be updated to reflect the new field.

company/models.py
monikasulik: Also, not sure what you discussed with @konradko but is this definitely the right decision (rather than having the company FK on user or an m2m which I think was mentioned on slack?) I don't know, just double checking and interested to hear you reasoning :)

monikasulik: It's just that I overheard a conversation between Lily and @konradko yesterday about there being migration of existing companies from other systems for alpha. And Konrad confidently said there would be a migration, which is why I'm assuming it's a requirement. But I may have misunderstood. Konrad would need to clear this up.

The reason I'm asking why user on the company rather than company on user is because if one user has many companies that export stuff, they are fairly likely to have separate emails for each of their companies and can therefore register more than once. Not ideal, but possible. If, on the other hand, multiple employees of a company needed access to the company profile on the site, we have no way of providing them with it with this modelling.

PR #12, opened 2016-10-18
DRF view for user retrieve and update

monikasulik:
https://uktrade.atlassian.net/browse/ED-155


company/views.py
```diff
@@ -7,7 +7,5 @@
 class CompanyRetrieveUpdateAPIView(RetrieveUpdateAPIView):
 
     model = Company
+    queryset = model.objects
```
richtier: does this work ok? I thought this would have to be `model.objects.all()`. better yet more explicit `Company.objects.all()` so we can read it all in one line :)

Similar comment for `UserRetrieveUpdateAPIView`.

monikasulik: Yep, works fine (according to tests at least :D). `get_queryset` adds an `all()` to it.
I think `model.objects.all()` is actually the risky one because the statement might get evaluated when the server starts the application instead of when the view is accessed. But if preferable I can just do:
```
def get_queryset(self):
    return self.model.objects.all()
```
The one thing I can't do is remove the queryset thing all together cause then the views don't work (according to my tests anyway).

monikasulik: Ah, actually I needn't have worried about the statement evaluation. This is what the generic view actually does:
```
    if isinstance(queryset, QuerySet):
            # Ensure queryset is re-evaluated on each request.
            queryset = queryset.all()
```
Though I'd still argue it's better not to add the `.all()` on the attribute because it will get evaluated when the application gets started and we'll have a db hit for absolutely no good reason.

richtier: django querysets are [lazily evaluated](https://docs.djangoproject.com/en/1.10/topics/db/queries/#querysets-are-lazy), so there is no danger of doing `Company.object.all()`  there :)

monikasulik: To my knowledge - yes they're evaluated lazily, but only in whatever scope they're in. In this case, the scope is the view class rather than any of its methods. I've seen at least one bug when somebody called all() there in a regular django view and the application returned the queryset as it was when the server was started.
<a href="http://stackoverflow.com/questions/19707237/use-get-queryset-method-or-set-queryset-variable">This stackoverflow response</a> corroborates my view on the matter ;) Although the django docs actually don't, which makes me think this may have been an issue in past versions but no longer.
Either way, I would err on the side of caution and setting managers instead of querysets was supported even in the days of generic function-based views. If you're uncomfortable with that I can happily go back to using just the `get_queryset` method - that's both clear and definitely safe ;)

konradko: DRF docs http://www.django-rest-framework.org/api-guide/generic-views/ are using ``queryset = User.objects.all()``, so we can consider it safe :-)

Also ``model`` is deprecated - only ``queryset`` needs to be specified, so please remove it.

user/views.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework.generics import RetrieveUpdateAPIView
+
+from user.serializers import UserSerializer
+from user.models import User
+
+
+class UserRetrieveUpdateAPIView(RetrieveUpdateAPIView):
+
+    model = User
```
konradko: ditto, please remove and use ``queryset = User.objects.all()``

monikasulik:
@konradko @richtier Have acted on comments - please rereview :)

PR #13, opened 2016-10-18
User Manager & User.is_superuser field

monikasulik:
Django management commands (create_superuser etc.) don't work without this.


user/models.py
```diff
@@ -1,9 +1,38 @@
...
 from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
 
 
+class UserManager(BaseUserManager):
+    use_in_migrations = True
+
+    def _create_user(self, email, password, **extra_fields):
+        if not email:
+            raise ValueError('The given email must be set')
```
richtier: feels like an `assert email, 'The given email must be set'` might be preferable.

monikasulik:
Tests added

PR #14, opened 2016-10-19
Add company email address confirmation endpoint

konradko:
https://uktrade.atlassian.net/browse/ED-230


api/settings.py
```diff
@@ -198,7 +196,7 @@
 # Auth
 
 AUTH_USER_MODEL = 'user.User'
-ACCOUNT_USER_MODEL_USERNAME_FIELD = 'email'
+ACCOUNT_USER_MODEL_USERNAME_FIELD = 'company_email'
```
konradko: this is gonna change with SSO integration, fine to stay like this for now

user/views.py
```diff
@@ -0,0 +1,35 @@
+import json
+
+from rest_framework import status
+from rest_framework.generics import GenericAPIView
+from rest_framework.response import Response
+
+
+class ConfirmCompanyEmailAPIView(GenericAPIView):
+
+    http_method_names = ("get", )
```
monikasulik: I don't think this should be a GET view - GET should always be indempotent: http://restcookbook.com/HTTP%20Methods/idempotency/

enrolment/queue.py
```diff
@@ -179,75 +186,49 @@ def create_objects(self, json_payload, message_id):
...
-            logger.error(
-                'Cannot create company. Invalid details.',
-                extra={'errors': serializer.errors}
-            )
-            raise  # trigger transaction rollback in parent function.
-        else:
-            serializer.save()
+
+        serializer.is_valid(raise_exception=True)
+        return serializer.save()
```
richtier: :+1:

PR #15, opened 2016-10-19
Link User to Company

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-228)


konradko:
@richtier needs updating with 'company_email' instead of 'email'

PR #21, opened 2016-10-20
Add health-check endpoint

konradko:


api/tests/test_views.py
```diff
@@ -0,0 +1,17 @@
+import json
+
+from django.core.urlresolvers import reverse
```
monikasulik: This actually is deprecated as of Django 1.10, but I guess in this repo we're on 1.9?

konradko: yup

PR #24, opened 2016-10-21
Reinstate Enrolment as a persistence tool

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-250)

The purpose of `Enrolment` is to make messages from the queue persistent to allow us to react to problems later (e.g., for some reason the user did not get created).


user/models.py
```diff
@@ -97,7 +97,7 @@ class User(AbstractBaseUser):
         ),
     )
 
-    company = models.ForeignKey(Company, related_name='users')
+    company = models.ForeignKey(Company, related_name='users', null=True)
```
konradko: why ``null=True``? Creation of user and company should still happen within a transaction.

enrolment/queue.py
```diff
@@ -112,7 +109,7 @@ def process_message(self, message):
                     json_payload=message.body,
                     sqs_message_id=message.message_id,
                 )
-            except ValidationError:
+            except (ValidationError, KeyError):
```
konradko: ``KeyError`` is too generic, please check where necessary that the key is there and raise ValidationError if not

enrolment/queue.py
```diff
@@ -93,8 +91,7 @@ def is_valid_enrolment(message_body):
         except ValueError:  # includes JSONDecodeError
             return False
         else:
-            serializer = serializers.EnrolmentSerializer(data=enrolment)
```
konradko: We should still use serializer, just with a 'data' field

enrolment/queue.py
```diff
@@ -149,86 +146,20 @@ def is_postgres_unique_violation_error(exception):
...
+    def process_enrolment(self, sqs_message_id, json_payload):
+        """Persis the message in enrolment.models.Enrolment
 
         Args:
             sqs_message_id (str): SQS message ID
-            aims (str[]): Goals of joining the scheme
-            company_number (str): Companies House number
-            company_email (str): User's company email
-            personal_name (str): User's full name
+            json_payload (str[]): Goals of joining the scheme
```
konradko: does not look like correct description

enrolment/serializers.py
```diff
@@ -1,19 +1,51 @@
...
 
 from enrolment import models
+from company.serializers import CompanySerializer
+from user.serializers import UserSerializer
 
 
 class EnrolmentSerializer(serializers.ModelSerializer):
 
     id = serializers.CharField(read_only=True)
+    data = serializers.JSONField(binary=True)
```
konradko: binary?

enrolment/tests/test_queue.py
```diff
@@ -61,101 +58,3 @@ def test_process_message_creates_company(self):
         assert instance.aims == VALID_REQUEST_DATA['aims']
         assert instance.number == VALID_REQUEST_DATA['company_number']
 
-    @pytest.mark.django_db
```
konradko: what happened to these tests? The validation errors will still be raised right? We should keep them, just with updated assumptions

user/models.py
richtier: what about creating users from the shell or `createsuperuser` command?

enrolment/serializers.py
richtier: converts the json string into a python object.

enrolment/queue.py
```diff
@@ -88,13 +86,10 @@ def is_valid_enrolment(message_body):
...
             boolean: True if valid, False if not
         """
-        try:
-            enrolment = json.loads(message_body)
-        except ValueError:  # includes JSONDecodeError
-            return False
-        else:
-            serializer = serializers.EnrolmentSerializer(data=enrolment)
-            return serializer.is_valid()
+        serializer = serializers.EnrolmentSerializer(
```
richtier: `JSONField` also checks if `message_body` is valid json for us, so no need for the above try except.

PR #26, opened 2016-10-21
Add check for if company exists and is active

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-261)

We will also need to change `directory-api-client` and `directory-ui` to treat the endpoint as a generic "validate the company number" rather than a "validate the uniqueness of the company number".


company/helpers.py
```diff
@@ -0,0 +1,24 @@
...
+import logging
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+from django.conf import settings
+
+
+logger = logging.getLogger(__name__)
+company_profile_url = 'https://api.companieshouse.gov.uk/company/{number}'
```
monikasulik: I know it's highly unlikely this will ever change, but to me it's the sort of thing that feels like it belongs in settings.py

company/serializers.py
```diff
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
-from company import models
+from company import models, validators
```
monikasulik: Shouldn't these be in the validators repo if we're going with the separate repo route?

company/validators.py
```diff
@@ -0,0 +1,19 @@
+import http
+
+from rest_framework.serializers import ValidationError
+
+from company.models import Company
+from company import helpers
+
+
+def company_unique(number):
+    if Company.objects.filter(number=number).exists():
```
monikasulik: Shouldn't this be ```number.strip()``` as a precaution?

company/validators.py
```diff
@@ -0,0 +1,19 @@
...
+        raise ValidationError('Already registered')
+
+
+def company_active(number):
+    response = helpers.get_companies_house_profile(number)
```
monikasulik: ditto

company/helpers.py
richtier: I agree, its not going to change.

company/validators.py
richtier: the serializer field strips automatically

company/helpers.py
monikasulik: Fair enough

company/serializers.py
monikasulik: I am definitely in favour of dropping directory-validators 🎉 Please please lets do that 🍰 I didn't like that idea from the start.

PR #28, opened 2016-10-23
Send email with confirmation url

monikasulik:
https://uktrade.atlassian.net/browse/ED-143
Not all the stuff that needs to be done as part of ED-143 is here. AFAIK, still need to do:
* save when email gets confirmed
* let directory-ui know that email has been confirmed

Also, this PR is using a test template atm.


monikasulik:
@konradko Atm CircleCI is failing because the API and service keys are not in the env variables. Where should these be added and what should be added? (as I'm mocking stuff, we can probably put in completely fake ones)

enrolment/queue.py
```diff
@@ -233,3 +242,16 @@ def save_company(self, aims, number):
 
         serializer.is_valid(raise_exception=True)
         return serializer.save()
+
+    def send_confirmation_email(self, user):
+        service_id = settings.NOTIFY_SERVICE_ID
+        api_key = settings.NOTIFY_API_KEY
+        template_id = settings.CONFIRMATION_EMAIL_TEMPLATE_ID
+        user.confirmation_code = get_random_string(64).lower()
```
monikasulik: There were many potential places where this could be done. I can explain why I decided to do it here if there are questions about this.
Possibly should check whether confirmation_code is already filled in so this can be used for resending also?

enrolment/queue.py
```diff
@@ -177,13 +179,20 @@ def process_enrolment(self, json_payload, sqs_message_id):
...
-        self.save_user(
+        user = self.save_user(
             company_email=payload['company_email'],
             name=payload['personal_name'],
             referrer=payload['referrer'],
             plaintext_password=payload['password'],
             company=company,
         )
+        # If there's an exception during email sending, the db
+        # transaction should complete and the exception should be logged
```
monikasulik: Is this the correct behaviour? This is what I assumed was probably best, but definitely worth verifying.

api/settings.py
```diff
@@ -182,3 +182,16 @@
 
 AUTH_USER_MODEL = 'user.User'
 ACCOUNT_USER_MODEL_USERNAME_FIELD = 'company_email'
+
+# Settings for Gov UK Notify
+# NOTE: The service id & api key can be generated at
+# https://www.notifications.service.gov.uk
+# in the API Integration > API Keys section
+
+NOTIFY_SERVICE_ID = os.getenv("GOV_UK_NOTIFY_SERVICE_ID")
```
konradko: easier to have same name for both the django setting and the env var, e.g.

```
GOV_NOTIFY_SERVICE_ID = os.getenv("GOV_NOTIFY_SERVICE_ID")
```

api/settings.py
```diff
@@ -182,3 +182,16 @@
...
+NOTIFY_API_KEY = os.getenv("GOV_UK_NOTIFY_API_KEY")
+# TODO: This is a temporary test template. Needs to be changed.
+# Template must contain ((confirmation url)) variable
+CONFIRMATION_EMAIL_TEMPLATE_ID = "db881d48-9eba-4401-acea-ebc1609db31e"
```
konradko: should also be retrieved from settings

api/settings.py
```diff
@@ -182,3 +182,16 @@
...
 AUTH_USER_MODEL = 'user.User'
 ACCOUNT_USER_MODEL_USERNAME_FIELD = 'company_email'
+
+# Settings for Gov UK Notify
+# NOTE: The service id & api key can be generated at
+# https://www.notifications.service.gov.uk
+# in the API Integration > API Keys section
+
+NOTIFY_SERVICE_ID = os.getenv("GOV_UK_NOTIFY_SERVICE_ID")
+NOTIFY_API_KEY = os.getenv("GOV_UK_NOTIFY_API_KEY")
```
konradko: ditto

api/settings.py
```diff
@@ -182,3 +182,16 @@
...
+# TODO: This is a temporary test template. Needs to be changed.
+# Template must contain ((confirmation url)) variable
+CONFIRMATION_EMAIL_TEMPLATE_ID = "db881d48-9eba-4401-acea-ebc1609db31e"
+# TODO: What is the correct url?
+CONFIRMATION_URL_TEMPLATE = "http://localhost/confirm?token=%s"
```
konradko: same, should be from env vars

enrolment/queue.py
konradko: would be better for the ogic of user / company creation to reside in ``enrolment.models`` (and connect them to post-save signal of the ``Enrolment`` model) instead of the queue module.

As for ``send_confirmation_email``, it should not be part of the transaction - better to hook it up to the post-save signal of ``User``.

enrolment/queue.py
konradko: we need a unique identifier, so better to use ``str(uuid.uuid4())``, we can also include user id in the confirmation URL (UUID is 36 chars, so we can concatenate it).

enrolment/queue.py
konradko: mind that was actually moved to serializers by @richtier in https://github.com/uktrade/directory-api/pull/24 - conflicts

api/settings.py
monikasulik: But I presume we do want a default or no? I should probably ticket this, so that a proper template (rather than my testing one) is created.

api/settings.py
monikasulik: Is there a default I should use?

enrolment/queue.py
monikasulik: I did actually think very seriously about adding the creation of the confirmation code and sending of the email to the user model. In the end decided not to because I could imagine all kinds of edge cases where maybe we wouldn't want this to happen (adding superusers for django's admin for example or having to manually change something/add a user because a site user is having problems etc.). I also thought that email sending would probably be something that we would want to do in a queue if we're worried about spikes?
How do you feel about me putting the whole confirmation code/email send logic into the serializer as per what's in @richtier 's PR? To me I think that seems the most sensible option atm (but I could well still change my mind ;)).

enrolment/queue.py
monikasulik: The only reason I went with a 64 char random string using django utils was that that's what allauth does :) Figured they probably knew what they were doing. But I'm happy to do uuid - makes sense to me. Don't think there's any point in having it longer than 36 chars or is there?

api/settings.py
```diff
@@ -70,7 +70,9 @@
 # https://docs.djangoproject.com/en/1.9/ref/settings/#databases
 
 DATABASES = {
-    'default': dj_database_url.config()
+    'default': dj_database_url.config(
+        default="postgres://test:test@localhost:5432/directory-test"
```
richtier: I prefer removing this and instead requiring it to be specified in env variables.

Motivation is without `default` - if the docker command running the app was to miss the `DATABASE_URL` then the app would fail with obvious error much easier to debug. If it defaulted to test db it would be more difficult to debug,

enrolment/queue.py
richtier: that logic now lives inside the Enrolment serializer.

api/settings.py
```diff
@@ -184,3 +184,8 @@
 ACCOUNT_USER_MODEL_USERNAME_FIELD = 'company_email'
 
 COMPANIES_HOUSE_API_KEY = os.getenv('COMPANIES_HOUSE_API_KEY')
+
+# Settings for Confirmation Emails
+CONFIRMATION_EMAIL_SUBJECT = 'Confirm your email address'
```
konradko: should also be an env var

enrolment/queue.py
```diff
@@ -144,6 +144,13 @@ def is_postgres_unique_violation_error(exception):
 
     def process_enrolment(self, sqs_message_id, json_payload):
         """Persis the message in enrolment.models.Enrolment
+        # If there's an exception during email sending, the db
```
konradko: this logic should not reside in the queue worker, models / serializers are a better place

enrolment/queue.py
```diff
@@ -144,6 +144,13 @@ def is_postgres_unique_violation_error(exception):
...
+        # transaction should complete and the exception should be logged
+        try:
+            self.send_confirmation_email(user)
+        except:
```
konradko: should catch specific errors

makefile
```diff
@@ -119,8 +121,10 @@ DEBUG_CREATE_DB := \
 debug_db:
 	$(DEBUG_SET_ENV_VARS) && $(DEBUG_CREATE_DB)
 
+PYTEST_DEBUG := pytest . --cov=. -s $(pytest_args)
+
 debug_test:
-	$(DEBUG_SET_ENV_VARS) && $(PYTEST)
+	$(DEBUG_SET_ENV_VARS) && $(PYTEST_DEBUG)
```
konradko: this looks same as ``PYTEST`` - the reason for using one var is for circleci tests and local tests to use the same command

templates/confirmation_email.html
```diff
@@ -0,0 +1 @@
+<a href="{{ confirmation_url }}">Confirmation url</a>
```
konradko: We need a ticket for creating proper template

user/signals.py
```diff
@@ -0,0 +1,27 @@
+import uuid
+
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives
+from django.template.loader import render_to_string
+
+
+def send_confirmation_email(sender, instance, created, *args, **kwargs):
+    if not created or not instance.company_email:
```
konradko: is this possible?

user/tests/test_signals.py
```diff
@@ -0,0 +1,71 @@
+import pytest
+from django.conf import settings
+from django.core import mail
+
+from user.models import User
+from user.signals import send_confirmation_email
+
+
+GOV_NOTIFY_EMAIL_METHOD = (
```
konradko: looks like a leftover

enrolment/queue.py
monikasulik: I think from a conversation we had earlier, when sending email we're not going to catch unexpected errors at all, so this code shouldn't be anywhere, much less so in the docstring of a method :)

makefile
monikasulik: I added the -s option in PYTEST_DEBUG cause I actually rather need it when debugging. On the other hand circleci shouldn't really have -s enabled cause it'll mean lots of unnecessary debug logging appearing etc. I can remove the whole thing if you prefer though.

api/settings.py
```diff
@@ -47,7 +47,7 @@
 TEMPLATES = [
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
-        'DIRS': [],
+        'DIRS': ['templates'],
```
konradko: shouldn't this be absolute path?

user/apps.py
```diff
@@ -1,5 +1,12 @@
...
 
+from django.db.models.signals import post_save
+
+from user.signals import send_confirmation_email
+
 
 class UserConfig(AppConfig):
     name = 'user'
+
+    def ready(self):
```
konradko: nice, finally a good place for signals

user/signals.py
konradko: ``company_email`` is nullable / allows blank?

user/signals.py
```diff
@@ -0,0 +1,27 @@
...
+        return
+
+    if not instance.confirmation_code:
+        instance.confirmation_code = str(uuid.uuid4())
+        instance.save()
+
+    subject = settings.CONFIRMATION_EMAIL_SUBJECT
+    from_email = settings.CONFIRMATION_EMAIL_FROM
+    url = settings.CONFIRMATION_URL_TEMPLATE % instance.confirmation_code
```
konradko: ``.format(confirmation_code=instance.confirmation_code)`` would be more explicit

user/signals.py
```diff
@@ -0,0 +1,27 @@
...
+
+    text_body = render_to_string('confirmation_email.txt',
```
konradko: kwargs are nice :P

templates/confirmation_email.html
konradko: 👍

user/signals.py
monikasulik: Nope, company_email is not nullable and does not allow blank. But allowing or not allowing blank is just a validation thing (possible to bypass) and not null does not prohibit empty strings.

konradko: true

api/settings.py
monikasulik: Tests claim that it works as it is, so I assumed this was fine?

user/signals.py
monikasulik: I can put the context into a separate variable so it's DRY, but can't actually use it as kwargs here ;) But I might be misunderstanding something.

PR #31, opened 2016-10-24
Enrolment field changes

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-263)

We have vastly different fields now - company revenue, size, user phone number, etc.


company/tests/test_models.py
```diff
@@ -15,8 +15,7 @@ def test_company_model_str():
 
 @pytest.mark.django_db
 def test_company_model_sets_string_fields_to_empty_by_default():
-    company = Company.objects.create(
-        number="01234567", aims=['AIM1', 'AIM2'])
+    company = Company.objects.create(number="01234567", revenue=10000)
```
monikasulik: Unlike aims was previously, revenue is actually a nullable field, so I don't think it should be added everywhere in tests (I totally support it being added to VALID_REQUEST_DATA, but think that elsewhere it's actually confusing cause, at least to me, it gives the impression that it's either required or somehow effects the test)

company/tests/test_serializers.py
```diff
@@ -32,37 +39,47 @@ def test_company_serializer_doesnt_accept_number_over_8_chars():
...
+    assert serializer.is_valid(), serializer.errors
 
     company = serializer.save()
 
     # NOTE: These are fields that the registration form currently
     # doesn't define and therefore might be empty.
     # This test is just for peace of mind that we handle this in a
     # consistent manner
-    assert company.name == ''
+    assert company.name == 'Extreme corp'
```
monikasulik: This test is here to check the combo of `default='', null=True, blank=True` brings expected results. Since this combo is no longer present on company.name, IMO this assert is confusing and not really relevant for this test.

company/tests/test_serializers.py
```diff
@@ -32,37 +39,47 @@ def test_company_serializer_doesnt_accept_number_over_8_chars():
...
+    assert serializer.is_valid(), serializer.errors
 
     company = serializer.save()
 
     # NOTE: These are fields that the registration form currently
     # doesn't define and therefore might be empty.
     # This test is just for peace of mind that we handle this in a
     # consistent manner
-    assert company.name == ''
+    assert company.name == 'extreme corp'
```
monikasulik: ditto

company/tests/test_serializers.py
```diff
@@ -78,7 +95,6 @@ def test_company_serializer_save():
     assert company.number == VALID_REQUEST_DATA['number']
     assert company.website == VALID_REQUEST_DATA['website']
     assert company.description == VALID_REQUEST_DATA['description']
-    assert company.aims == VALID_REQUEST_DATA['aims']
```
monikasulik: This is testing that all relevant fields are getting saved correctly, so IMO should have revenue, export_status etc. added in the asserts here.

company/tests/test_serializers.py
```diff
@@ -13,13 +15,18 @@ def test_company_serializer_doesnt_accept_number_under_8_chars():
```
monikasulik: aims is still in this test

user/tests/test_serializers.py
```diff
@@ -52,11 +54,12 @@ def test_user_serializer_save():
     assert user.is_staff is False
     assert user.password == ''
     assert user.last_login is None
+    assert user.mobile_number == VALID_REQUEST_DATA['mobile_number']
 
 
 @pytest.mark.django_db
 def test_user_with_company_serializer_save():
-    company = Company.objects.create(aims=['1'])
+    company = Company.objects.create(revenue=20000)
```
monikasulik: I don't think the revenue field needs to be here (as it's nullable).

company/models.py
```diff
@@ -10,17 +11,37 @@ class Company(models.Model):
...
-        validators=[MinLengthValidator(8)],
+        validators=[shared_validators.company_number],
         unique=True
     )
-    website = models.URLField(max_length=255, blank=True, null=True,
-                              default='')
-    description = models.TextField(blank=True, null=True, default='')
-    aims = JSONField()
-    logo = models.FileField(blank=True, null=True)
+    revenue = models.DecimalField(
```
richtier: Django docs suggest using decimal for money. Worst case scenario is there are unwanted decimal places here :)

company/tests/test_models.py
richtier: yeah I made the following mistake:
1) implemented the field. Made it non-nullable
2) updated the tests to include revenue
3) make the field nullable
4) oops tests now have revenue

will sort that out

company/models.py
konradko: 👍

company/tests/test_serializers.py
richtier: done :)

PR #40, opened 2016-10-28
Use SSO, squash migrations, fix company view to retrieve via sso_user_id

konradko:


company/views.py
```diff
@@ -39,5 +39,10 @@ def get(self, request, *args, **kwargs):
 
 class CompanyRetrieveUpdateAPIView(RetrieveUpdateAPIView):
 
-    queryset = Company.objects.all()
     serializer_class = CompanySerializer
+
+    def get_object(self):
+
+        return User.objects.get(
```
monikasulik: Shouldn't this be somekind of get or 404 type thing? I mean if the sso_user_id that's passed is incorrect, atm this would return 500 which isn't great.

enrolment/tests/test_queue.py
```diff
@@ -75,18 +75,17 @@ def test_process_message_sends_confirmation_email(self):
 
         email = VALID_REQUEST_DATA['company_email']
         user = User.objects.get()
-        url = settings.CONFIRMATION_URL_TEMPLATE.format(
-            confirmation_code=user.confirmation_code)
+        url = settings.COMPANY_EMAIL_CONFIRMATION_URL_TEMPLATE.format(company_email_confirmation_code=user.company_email_confirmation_code)
```
monikasulik: flake8 is definitely not going to like you for this line :scream_cat:

company/views.py
```diff
@@ -39,5 +39,10 @@ def get(self, request, *args, **kwargs):
 
 class CompanyRetrieveUpdateAPIView(RetrieveUpdateAPIView):
 
-    queryset = Company.objects.all()
     serializer_class = CompanySerializer
+
+    def get_object(self):
```
monikasulik: This method should have somekind of test. Or at the very least (if we really don't have time) a ticket in the backlog to create this test.

company/views.py
konradko: this is automatic - get object will return 404 if DoesNotExist will be raised

company/views.py
konradko: it is tested: https://github.com/uktrade/directory-api/pull/40/files#diff-ab4b804cfb7c052c57aa42630c7940cfR39

company/views.py
konradko: sorry I am wrong - need to use get_or_404. Well spotted

PR #45, opened 2016-10-30
Add imagefield and django storages s3 for company logo

konradko:
image


company/utils.py
```diff
@@ -0,0 +1,16 @@
...
+
+def path_and_rename(path):
+    def wrapper(instance, filename):
+        ext = filename.split('.')[-1]
+        # get filename
+        if instance.pk:
+            filename = '{}.{}'.format(instance.pk, ext)
+        else:
+            # set filename as random string
+            filename = '{}.{}'.format(uuid4().hex, ext)
```
richtier: if the filename does not have a extension then it will end up like `sdfsdfsdfsdf.None`

PR #50, opened 2016-10-31
Add email uniqueness validator view

richtier:


api/urls.py
```diff
@@ -6,7 +6,11 @@
     CompanyNumberValidatorAPIView,
     CompanyRetrieveUpdateAPIView,
 )
-from user.views import UserRetrieveUpdateAPIView, ConfirmCompanyEmailAPIView
+from user.views import (
+    UserRetrieveUpdateAPIView,
+    ConfirmCompanyEmailAPIView,
+    UserEmailValidatorAPIView
```
konradko: would be best to call `UserCompanyEmailValidatorAPIView` to keep it consistent

api/urls.py
```diff
@@ -51,6 +55,11 @@
         name='validate-company-number'
     ),
     url(
+        r'validate-email-address/$',
```
richtier: I will do a follow up pr changing this and "validate company number" to be `/validators/x`

konradko: `/validate/x` will be fine

PR #57, opened 2016-11-06
Use only numbers in SMS code

konradko:


konradko:
@monikasulik added checking `code.isdigit()`

PR #59, opened 2016-11-08
Send invalid messages to separate queue

konradko:
https://uktrade.atlassian.net/browse/ED-375

### DIFF
--- a/enrolment/queue.py
+++ b/enrolment/queue.py
@@ -75,59 +75,6 @@ class Worker:
             gc.collect()
 
     @staticmethod
-    def is_valid_enrolment(message_body):
-        """Returns True if message body is valid models.Enrolment
-
-        Args:
-            message_body (SQS.Message.body): SQS message body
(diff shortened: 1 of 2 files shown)


enrolment/queue.py
```diff
@@ -99,64 +99,46 @@ def process_message(self, message):
...
+                logging.exception(
+                    "Failed to process message '{}'".format(
+                        message.message_id
+                    )
+                )
 
-    def process_enrolment(self, sqs_message_id, json_payload):
-        """Persist the message in enrolment.models.Enrolment
+        # Delete the message only if it was sucessfully processed or caught by
+        # the exceptions above, as otherwise we want to leave it in the queue
```
konradko: yes, this is what we want, e.g. if a bug was deployed / particular worker instance got spoiled somehow. If it was saved to db and then errored, new objects won't be created, cause of unique sqs ID constraint

PR #67, opened 2016-11-18
Expose year of incorporation

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-350)


company/management/commands/retrieve_missing_company_details.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Command(BaseCommand):
+    help = 'Retrieves missing data of companies such as date of creation'
+
+    def handle(self, *args, **options):
+        for company in Company.objects.filter(date_of_creation__isnull=True):
+            date_of_creation = helpers.get_date_of_creation(company.number)
+            company.date_of_creation = date_of_creation
+            company.save()
```
richtier: We should run this on prod some time. Note sure how we can run it though. @konradko?

conftest.py
```diff
@@ -0,0 +1,26 @@
+import http
```
richtier: this file allows us to set mocks that will be used for every test.

company/management/commands/retrieve_missing_company_details.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Command(BaseCommand):
+    help = 'Retrieves missing data of companies such as date of creation'
+
+    def handle(self, *args, **options):
+        querystring = models.Company.objects.exclude(number='')
+        for company in querystring.filter(date_of_creation__isnull=True):
+            date_of_creation = helpers.get_date_of_creation(company.number)
+            company.date_of_creation = date_of_creation
+            company.save()
```
richtier: We should run this on prod some time so we retrieve the date of creation for existing companies. Not sure how we can run it though. @konradko?

company/models.py
```diff
@@ -55,6 +55,13 @@ class Company(models.Model):
         max_length=255, blank=True, null=True,
         default=''
     )
+    date_of_creation = models.DateField(blank=True, null=True, default=None)
 
     def __str__(self):
         return self.name
+
+    @helpers.continue_on_network_error
```
richtier: Added this because we don't want the company enrolment to fail if companies house is down, or any of [the possible exceptions](https://github.com/uktrade/directory-api/pull/67/files#diff-fd9c4c1227c38dd2870187b612436f14R48) occur.

Note we can run `manage.py retrieve_missing_company_details` periodically to counteract errors that prevent the date of creation being pulled down.

conftest.py
```diff
@@ -0,0 +1,26 @@
...
+        'GET',
+        re.compile('https://api.companieshouse.gov.uk/company/.*'),
+        json=company_profile,
+        status_code=http.client.OK,
+    )
+
+    helpers.companies_house_session.mount(
+        'https://api.companieshouse.gov.uk',
+        companies_house_adapter
+    )
```
richtier: this is a rad feature of `requests` - it allows us to easily mock responses for all tests.

company/management/commands/retrieve_missing_company_details.py
konradko: we can run one-off tasks with overwritten commands via ECS

company/models.py
```diff
@@ -55,6 +55,13 @@ class Company(models.Model):
...
+    def save(self, *args, **kwargs):
+        if not self.id:
+            self.date_of_creation = helpers.get_date_of_creation(self.number)
```
konradko: this will be the 2nd call to companies house that we do during enrolment, while our API rate limit is pretty low. Can we take the creation date in UI, together with the name? And then just add a new field to `Enrolment`.

richtier: so confirm change required, are we going for:

 - cache CH calls; or
 - move CH calls to directory-ui

?

PR #69, opened 2016-11-23
Add admin

konradko:


richtier:
how will `EnrolmentAdmin` be assessed? we don't expose api to the internet.

PR #70, opened 2016-11-24
Supplier case study

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-274)

This PR adds the model, view, and serializer needed for the supplier case study.

This PR also removes the obsolete 'companies house profile' endpoint.


company/models.py
```diff
@@ -59,3 +59,52 @@ class Company(models.Model):
...
+    image_one = models.FileField(
+        null=True,
+        blank=True,
+        default='',
+        upload_to=helpers.path_and_rename_supplier_case_study,
+    )
+    image_two = models.FileField(
+        null=True,
+        blank=True,
+        default='',
```
monikasulik: Is default='' definitely the behaviour we want for file fields? Not sure they follow the same pattern as django's char-based fields.

company/tests/test_serializers.py
```diff
@@ -97,6 +130,21 @@ def test_company_serializer_save():
     assert company.export_status == VALID_REQUEST_DATA['export_status']
 
 
+@pytest.mark.django_db
+def test_company_serializer_nested_case_stody(
```
monikasulik: typo: study not stody :)

richtier: I'm writing with an accent haha. Will fix.

company/tests/test_models.py
```diff
@@ -30,3 +32,21 @@ def test_company_enforces_unique_company_number():
     models.Company.objects.create(number="01234567")
     with pytest.raises(IntegrityError):
         models.Company.objects.create(number="01234567")
+
+
+@pytest.mark.django_db
```
monikasulik: Actually, I was hoping for a bit more testing. See: https://github.com/uktrade/directory-api/commit/1c484987ee99f1707116f6e8c102e5a0a3ebd87a I know we already have tests for other models/fields so we know that solution works, but it's a good idea just in case somebody decides to run havoc on the code long after we're gone ;)
The purpose there is that we want to be sure that no matter what we give the model/serializer (None or '') it will behave the same way, so we don't have both NULL and '' in the db as per django's guidelines. And that was the only reason for going the insane `blank=True, null=True, default=''` route cause in normal circumstances `blank=True` would have been enough.

richtier: cool, will add

PR #71, opened 2016-11-25
Add published flag on company

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-480)

image


company/models.py
```diff
@@ -56,6 +56,7 @@ class Company(models.Model):
         default=''
     )
     date_of_creation = models.DateField(blank=True, null=True, default=None)
+    is_published = models.BooleanField(default=False)
```
monikasulik: Possibly serializer tests should be updated to check this is False on save?

PR #72, opened 2016-11-25
Generating csv of users and companies

monikasulik:
TODO:

- [x] Unit test using data for user/company that has all fields filled in
- [x] Unit test for csv with multiple lines

Actually integrating with the admin will be in another PR.


user/tests/test_admin.py
```diff
@@ -0,0 +1,133 @@
...
+    'referrer', 'is_active', 'date_joined', 'terms_agreed',
+    'company_id', 'company_name', 'description', 'employees',
+    'export_status', 'keywords', 'logo', 'number', 'revenue', 'sectors',
+    'website', 'date_of_creation', 'is_published']
+
+
+@pytest.fixture
+def company():
+    # fill in optional fields as well
+    # TODO: logo should not be None. See ED-489
```
konradko: why not just set it to some random string?

monikasulik: I wanted to use an actual django-like file object, to check that the path to the file was actually being dumped into the csv.

user/tests/test_admin.py
```diff
@@ -0,0 +1,133 @@
...
+from user.models import User
+from user.tests import VALID_REQUEST_DATA as USER_DATA
+from user.admin import export_users_to_csv
+from company.models import Company
+from company.tests import VALID_REQUEST_DATA as COMPANY_DATA
+
+
+EXPECTED_CSV_FIELDS = [
+    'id', 'sso_id', 'name', 'mobile_number', 'company_email',
+    'company_email_confirmed', 'company_email_confirmation_code',
```
monikasulik: There's probably a few fields that aren't too relevant there, but I figured I'd dump them all as it doesn't hurt. Personally, I'd leave it unless there's a good reason (security, functional etc.) to remove it, but your call.

user/admin.py
```diff
@@ -1,6 +1,42 @@
...
+    company_fieldnames += [field.name for field in Company._meta.fields
+                           if field.name not in ['id', 'name']]
+    fieldnames = user_fieldnames + company_fieldnames
+
+    # Retrieve company data straight away to avoid extra calls
+    queryset = User.objects.select_related('company').all()
+
+    writer = csv.DictWriter(csv_output, fieldnames=fieldnames)
+    writer.writeheader()
+    for user in queryset:
```
monikasulik: There are two potential issues with this IMO. One is that the user object has fields like `company_email`, which looks quite similar to the `company__*` notation and could be confusing (this csv is for people who do not know the django notation :)). The other is that when new company or user fields are added, people are likely to forget to update the csv generation. Atm what would happen on a new field being added is that the csv tests would fail and the person who added the new field would have to think through whether they want that field to be in the csv or not and update either the tests or the code. In your version this would not happen and therefore the csv code is likely not to get updated.

konradko: the fields we include should be curated, even just in case we will ever keep any secrets there

user/tests/test_admin.py
konradko: including all by default feels more risky from sec perspective

user/tests/test_admin.py
konradko: the imagefield / filefield is just a string though

PR #73, opened 2016-11-25
Add public profile endpoint

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-481)

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -5,6 +5,7 @@ from api.views import documentation, HealthCheckAPIView
 from company.views import (
     CompanyCaseStudyViewSet,
     CompanyNumberValidatorAPIView,
+    CompanyPublicProfileRetrieveAPIView,
     CompanyRetrieveUpdateAPIView,
 )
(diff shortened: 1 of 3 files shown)


company/tests/test_views.py
```diff
@@ -211,6 +211,33 @@ def company():
...
+        revenue='100000.00',
+        is_published=True,
+    )
+
+
+@pytest.fixture
+def private_profile():
+    return Company.objects.create(
+        number="0123456B",
+        name='public company',
```
monikasulik: private company maybe for easier differentiation in ipdb and such? ;)

api/urls.py
```diff
@@ -70,6 +71,16 @@
         name='user'
     ),
     url(
+        r'company/public/(?P<companies_house_number>.*)/$',
```
konradko: the url arg pattern should match just letters and numbers (unless other chars can be in the CH num) - ATM it is too greedy.

PR #77, opened 2016-11-28
Filter companies by sector

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-502)

image

image

image

### DIFF
--- a/company/tests/test_views.py
+++ b/company/tests/test_views.py
@@ -22,6 +22,16 @@ from company.tests import (
 from user.models import User
 
 
+default_public_profile_data = {
+    'name': 'private company',
+    'website': 'http://example.com',
+    'description': 'Company description',
+    'export_status': choices.EXPORT_STATUSES[1][0],
+    'date_of_creation': '2010-10-10',
+    'revenue': '100000.00',
+}
+
+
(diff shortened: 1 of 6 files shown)


company/tests/test_views.py
```diff
@@ -419,3 +449,44 @@ def test_company_profile_public_list_profiles(
...
+@patch('signature.permissions.SignaturePermission.has_permission', Mock)
+def test_company_profile_public_list_profiles_empty_filter(
+    public_profile_software, public_profile_cars, public_profile_smart_cars,
+    api_client
+):
+    url = reverse('company-public-profile-list')
+
+    parsed = api_client.get(url).json()
+    parsed_ids = [int(item['id']) for item in parsed['results']]
+    assert parsed['count'] == 3, item
```
monikasulik: This `, item` thing looks a bit weird in the assert here. Is this meant to be a tuple? (as in `(3, item)`) And also, it does seem a little icky to be relying on `item` when it's actually a var declared in a list comprehension.

richtier: `, item` prints the value of `item` if the assertion fails. It was for debugging. will remove.

PR #78, opened 2016-11-29
Add create & update timestamps to all models

monikasulik:
https://uktrade.atlassian.net/browse/ED-499


user/models.py
```diff
@@ -24,7 +24,7 @@ class User(models.Model):
         unique=True,
     )
     company = models.ForeignKey(
-        Company, related_name='users', null=False
+        Company, related_name='users', null=True
```
monikasulik: Reason for change: https://uktrade.atlassian.net/browse/ED-512

richtier: I take it we're setting `null` to `True` so those 15 tests pass?

konradko: Incorrect / broken tests are not a good reason to model the data incorrectly. Users without a company are not allowed in the system, which has to be reflected on the DB level. We can do that in a separate ticket, given it used to be nullable and there were no migrations for it.

monikasulik: I already created a separate ticket to do this - it's at the start of this thread. Feel free to schedule it into this sprint if you feel it's urgent.

company/migrations/0015_auto_20161129_1621.py
```diff
@@ -0,0 +1,44 @@
+# -*- coding: utf-8 -*-
```
konradko: looks like a conflicting migration - some fields were added earlier (probably you just have to run migrate before generating the migration)

monikasulik: Should be fixed now

PR #82, opened 2016-12-01
Replace user with supplier

konradko:


company/views.py
```diff
@@ -20,8 +21,8 @@ class CompanyRetrieveUpdateAPIView(generics.RetrieveUpdateAPIView):
 
     def get_object(self):
         return generics.get_object_or_404(
-            models.Company, users__sso_id=self.kwargs['sso_id']
-        )
+            Supplier, sso_id=self.kwargs['sso_id']
+        ).company
```
richtier: not a fan of this change as it turns one lookup into 2. `get_object_or_404` is  not evaluated lazily.

konradko: true, non-optimal. Will make another PR to fix

PR #83, opened 2016-12-02
Optimize supplier company or 404 lookup

konradko:


company/helpers.py
```diff
@@ -76,3 +77,14 @@ def __call__(self, instance, filename):
 path_and_rename_supplier_case_study = PathAndRename(
     sub_path="/supplier_case_study"
 )
+
+
+def get_supplier_company_or_404(sso_id):
+    from supplier.models import Supplier
```
richtier: why is this not at the top of the module?

PR #84, opened 2016-12-02
Verification letters

konradko:
https://uktrade.atlassian.net/browse/ED-57


api/settings.py
```diff
@@ -226,3 +226,12 @@
...
 CSRF_COOKIE_SECURE = os.getenv('CSRF_COOKIE_SECURE') != 'false'
+
+# Verification letters sent with stannp.com
+FEATURE_VERIFICATION_LETTERS_ENABLED = os.getenv(
+    'FEATURE_VERIFICATION_LETTERS_ENABLED'
+) == 'true'
+
+STANNP_API_KEY = os.environ["STANNP_API_KEY"]
+STANNP_TEST_MODE = os.getenv('FEATURE_PUBLIC_PROFILES') != 'false'
+STANNP_VERIFICATION_LETTER_TEMPLATE_ID = os.environ["STANNP_TEST_MODE"]
```
richtier: `STANNP_TEST_MODE` and `STANNP_VERIFICATION_LETTER_TEMPLATE_ID` logic is a bit wrong :)

company/views.py
```diff
@@ -62,3 +64,49 @@ def get_serializer(self, *args, **kwargs):
...
+                suppliers__sso_id=serializer.data[
+                    'sso_user_id'
+                ]
+            )
+        except models.Company.DoesNotExist:
+            response_status_code = status.HTTP_400_BAD_REQUEST
+            response_data = json.dumps({
+                "status_code": response_status_code,
+                "detail": "Company not found"
+            })
```
richtier: could be simplified a bit with [handle_exception](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/views.py#L417)

something like

```
   def handle_exception(self, exception):
        if isinstance(exception, models.Company.DoesNotExist):
              return Response(response_data, 400)
          return super().handle_exception(self, exception):

    def post(request, *args, **kwargs):
        company = models.Company.objects.get(
            suppliers__sso_id=serializer.data['sso_user_id']
         )
         ... # do rest of the stuff without try except block, because that code is handled by `handle_exception`

File: company/views.py [hunk 3]
```diff
@@ -62,3 +64,49 @@ def get_serializer(self, *args, **kwargs):
...
+        else:
+            if company.verification_code == serializer.data['code']:
```
richtier: this can be done in the serializer.

```
class VerifyCompanyWithCodeSerializer
    def validate_code(self, value):
        if value != self.context['expected_code']:
            # raise here
```

in `post`, we would then do
```
        serializer = self.serializer_class(
            data=request.data,
            context={'expected_code': company.verification_code}
        )
```

company/helpers.py
```diff
@@ -76,3 +76,17 @@ def __call__(self, instance, filename):
...
+def get_stannp_recipient(registered_address):
+    recipient = {}
+    recipient['title'] = ''
+    recipient['firstname'] = ''
+    recipient['lastname'] = ''
+    recipient['address1'] = registered_address['address_line_1']
+    recipient['address2'] = registered_address['address_line_2']
+    recipient['city'] = registered_address['locality']
+    recipient['postcode'] = registered_address['postal_code']
+    recipient['country'] = registered_address['country']
```
richtier: I think this should be hard-coded to UK

api/settings.py
konradko: whoops

company/views.py
konradko: 👍

company/helpers.py
konradko: I wouldn't hardcode it - if we require it to be UK, validation should be added

PR #86, opened 2016-12-07
Simplify enrolment process

richtier:


supplier/tests/test_admin.py
```diff
@@ -12,7 +12,7 @@
 
 headers = (
     'sso_id,name,mobile_number,company_email,company_email_confirmed,r'
-    'eferrer,is_active,date_joined,terms_agreed,company_id,company__na'
+    'eferrer,is_active,date_joined,company_id,company__na'
```
monikasulik: not at all important, but IMO would be better to have the r in this line rather than the previous one - a lot more readable and less likely to be corrected as a typo by somebody overzealous such as myself ;)

user/models.py
```diff
@@ -66,10 +52,12 @@ class User(TimeStampedModel):
         _('date joined'),
         default=timezone.now,
     )
-
-    terms_agreed = models.BooleanField(
-        _('terms_agreed'),
-        default=False
+    # depricated in favour of company.models.Company.contact_details
```
monikasulik: typo: should be deprecated :)

PR #96, opened 2016-12-12
Add additional testimonial fields

monikasulik:


company/models.py
```diff
@@ -116,7 +119,19 @@ class CompanyCaseStudy(TimeStampedModel):
     testimonial = models.CharField(
```
konradko: @richtier @monikasulik do we expect just a single testimonial? Feels like a separate `Testimonial` with FK to `CompanyCaseStudy` could be better here.

company/serializers.py
```diff
@@ -29,6 +32,15 @@ def validate_website(self, value):
     def validate_testimonial(self, value):
         return value or ''
 
+    def validate_testimonial_name(self, value):
+        return value or ''
```
konradko: Any reason for this? Seems a bit unnecessary, given that field value already defaults to ''. We allow `null` but never actually let it through.

company/tests/test_views.py
```diff
@@ -464,6 +475,11 @@ def test_company_case_study_get(
 
     assert response.status_code == http.client.OK
     assert data['testimonial'] == supplier_case_study.testimonial
+    assert data['testimonial_name'] == supplier_case_study.testimonial_name
+    assert data['testimonial_job_title'] == \
```
konradko: breaking lines with slash is discouraged, better to use parens:

```
assert data['testimonial_company'] == (
    supplier_case_study.testimonial_company
)
```

company/serializers.py
monikasulik: If you explicitly set something as None then django will interpret that as explicitly telling it to set the field to NULL. Without this tests which do this don't pass.

company/models.py
monikasulik: According to the doc that Rogier has shared - yep, we expect just a single testimonial. I'm going to have to slightly change the field names though as I didn't have that doc before and the names are slightly different.

konradko: Don't really like the change - before it was clear what the fields are related to, while `source` does not point to anything in the models. What we call these fields in the display layer is fine to be whatever is in the design / wireframes, for model readability I'd keep them to the names you choose originally - more explicit.

konradko: Other thing, would be good to check that single testimonial requirement with Alfie, as we might be unnecessarily putting limits here :)

monikasulik: Right, single testimonial requirement checked with Alfie and he's very confident that it's either 1 or 0 testimonials per case study and doesn't expect changes to that anytime soon.
I'll change the field names back and update uktrade/directory-ui/pull/262 accordingly as well.

PR #98, opened 2016-12-15
Populate company name on stannp envelope

richtier:
Currently the company name on the envelope is not being populated because we're not passing `company` to stannp.


richtier:
@monikasulik `company_name` is no longer needed, but we cant change the existing template to use `company` instead of `company_name` until `company` is being provided (letters that are being sent out in the time between the template updated and this PR being merged would not have `company`).

so my plan was to:
1) send both `company_name` and `company` to the template. get that merged in.
2) update the template to remove `company_name`, and instead use `company`.
3) stop sending `company_name` to the template.

That takes a bit longer but feels better practice for safety.

PR #100, opened 2016-12-19
Blacklist csv fields

monikasulik:
https://uktrade.atlassian.net/browse/ED-504


supplier/admin.py
```diff
@@ -47,11 +42,17 @@ def download_csv(self, request, queryset):
             )
         )
 
+        fieldnames = [field for field in Supplier._meta.get_all_field_names()
```
konradko: `get_all_field_names` is deprecated in Django 1.10 (https://docs.djangoproject.com/en/1.10/ref/models/meta/), please use the new way if possible

supplier/tests/test_admin.py
```diff
@@ -11,20 +15,21 @@
...
+    'company__logo,company__name,company__number,company__revenue,'
+    'company__sectors,company__website,company_email,'
+    'company_email_confirmed,company_id,date_joined,is_active,'
+    'mobile_number,name,sso_id'
 )
 
 
-class DownloadCSVTestCase:
+@pytest.mark.django_db
+class DownloadCSVTestCase(TestCase):
```
konradko: woah, good catch, lucky it ever worked

supplier/admin.py
```diff
@@ -13,27 +14,21 @@ class SupplierAdmin(admin.ModelAdmin):
...
-        'company__is_published'
+    csv_excluded_fields = (
+        'id',
+        'company',
+        'created',
+        'modified',
+        'company_email_confirmation_code',
+        'company__supplier_case_studies',
+        'company__suppliers',
+        'company__verified_with_code',
```
konradko: I think we want to include this field

supplier/admin.py
```diff
@@ -13,27 +14,21 @@ class SupplierAdmin(admin.ModelAdmin):
...
+        'company__is_verification_letter_sent',
```
konradko: ditto

supplier/admin.py
```diff
@@ -13,27 +14,21 @@ class SupplierAdmin(admin.ModelAdmin):
...
+        'company__contact_details',
```
konradko: this possibly too

supplier/admin.py
```diff
@@ -13,27 +14,21 @@ class SupplierAdmin(admin.ModelAdmin):
...
+        'company__verification_code',
+        'company__created',
```
konradko: ditto

supplier/admin.py
```diff
@@ -13,27 +14,21 @@ class SupplierAdmin(admin.ModelAdmin):
...
+        'company__modified',
```
konradko: ditto

konradko:
@monikasulik will the CSV include all the fields that are currently present after this change?

monikasulik:
Yep, it will include all fields from Supplier and Company unless explicitly excluded. Every time a field is added to either of those models the csv tests will fail (which means people will have to give it a thought about whether the field should or shouldn't be in the csv - I think that's good).
The fields are sorted alphabetically, so that there's some sort of consistency to the order in which they appear in the csv. Unfortunately control over field order is one thing we lose with this implementation.

konradko:
@monikasulik good thing with the test failing on model changes. It will be a bit annoying, but at least visible.

company/tests/test_models.py
```diff
@@ -65,7 +65,7 @@ def test_company_case_study_non_required_fields(company):
 
 
 def test_company_model_has_update_create_timestamps():
-    field_names = models.Company._meta.get_all_field_names()
+    field_names = [field.name for field in models.Company._meta.get_fields()]
```
konradko: 👍

supplier/tests/test_admin.py
```diff
@@ -1,30 +1,43 @@
...
+    'company__number,company__revenue,company__sectors,'
+    'company__twitter_url,company__verified_with_code,'
+    'company__website,company_email,company_email_confirmed,'
+    'date_joined,is_active,mobile_number,name,sso_id'
 )
 
 
-class DownloadCSVTestCase:
+COMPANY_DATA = VALID_REQUEST_DATA.copy()
+# Order in dict is unpredictable so for easier testing just 1 element
```
konradko: or you can use `OrderedDict` :P

monikasulik: True. Though as a lot of tests rely on this, can't be sure what the consequences would be. I need factory boy thingies ;)

PR #102, opened 2016-12-20
Add social media links to company

richtier:


company/serializers.py
```diff
@@ -88,6 +91,15 @@ def validate_website(self, value):
     def validate_description(self, value):
         return value or ''
 
+    def validate_twitter_url(self, value):
```
konradko: shouldn't this be checking that URL starts with twitter.com? Same below

konradko: can be another PR

PR #103, opened 2016-12-21
Improve email content for spam & consistency with sso

monikasulik:


api/templates/confirmation_email.html
```diff
@@ -1,12 +1,13 @@
...
+	{% load external_url from constants_tags %}
     <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
 	<html xmlns="http://www.w3.org/1999/xhtml">
 	<head>
 		<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
 		<meta name="viewport" content="width=device-width, initial-scale=1.0">
 		<meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
 		<meta name="format-detection" content="telephone=no" /> <!-- disable auto telephone linking in iOS -->
-		<title>Exporting is GREAT</title>
+		<title>Exporting is Great</title>
```
richtier: what is the intent here?

monikasulik: Consistency really. If I've removed GREAT completely from all sso emails (including title) then it makes sense to remove it from api too.
In any case, spam filters may be checking the whole html, not just the html body, so in theory this could still effect stuff (in practice I doubt one instance in title would, but...).

PR #106, opened 2016-12-23
Publishing companies by giving a list of company house numbers

monikasulik:
https://uktrade.atlassian.net/browse/ED-571


company/admin.py
```diff
@@ -1,12 +1,64 @@
 from django.contrib import admin
+from django.conf.urls import url
+from django.views.generic import FormView
+from django.core.urlresolvers import reverse_lazy
+from django import forms
 
 from company.models import Company, CompanyCaseStudy
 
 
+class CHNumbersForm(forms.Form):
```
richtier: we're not rationing letters :) please call it `CompaniesHouseNumbersForm`

company/admin.py
```diff
@@ -1,12 +1,64 @@
...
+    COMPANY_DOESNT_EXIST_MSG = (
+        'Some companies in this data set are not in the db'
+    )
+
+    company_numbers = forms.CharField(
+        widget=forms.Textarea,
+        help_text='Comma-separated company house numbers'
```
richtier: if users are copying from a spreadsheet then it would be much more user friendly to separate by line termination rather than comma.

company/admin.py
```diff
@@ -1,12 +1,64 @@
...
+    )
+
+    def clean_company_numbers(self):
+        ch_numbers = self.cleaned_data['company_numbers'].split(',')
+        ch_numbers = [num.strip() for num in ch_numbers if num.strip()]
```
richtier: `ch_` and `num` makes things less readable. Please consider `number` and `number`.

company/admin.py
```diff
@@ -1,12 +1,64 @@
...
+
+        number_of_companies = Company.objects.filter(
+            number__in=ch_numbers).count()
+        if number_of_companies != len(ch_numbers):
+            raise forms.ValidationError(self.COMPANY_DOESNT_EXIST_MSG)
```
richtier: this is not very helpful for the user - it does not tell them which company does not exist - meaning they cannot solve the problem.

I think this should instead be a warning instead - so perform the operation on all valid company numbers and in the response to the user say 'pssst! the following number was invalid!".

company/templates/admin/company/change_list.html
```diff
@@ -0,0 +1,22 @@
+{% extends "admin/change_list.html" %}
+{% load i18n admin_urls admin_static admin_list %}
+
+{% block object-tools %}
+  {% if has_add_permission %}
+    <ul class="object-tools">
+      {% block object-tools-items %}
+        <li>
+          {% url cl.opts|admin_urlname:'add' as add_url %}
```
richtier: what is `cl`?

company/templates/admin/company/publish_form.html
```diff
@@ -0,0 +1,36 @@
+{% extends "admin/change_form.html" %}
+{% load i18n admin_urls admin_static admin_modify %}
+
+{% block breadcrumbs %}
+<div class="breadcrumbs">
+<a href="{% url 'admin:index' %}">{% trans 'Home' %}</a>
+&rsaquo; <a href="{% url 'admin:app_list' app_label=opts.app_label %}">{{ opts.app_config.verbose_name }}</a>
```
richtier: `opts` is not very helpful for the reader. what can you give is a more helpful name please.

company/templates/admin/company/publish_form.html
```diff
@@ -0,0 +1,36 @@
...
+
+{% block content %}
+<div id="content-main">
+	<form action="" method="post">
+		{% csrf_token %}
+		{{ form.non_field_errors }}
+		<div>
+			<fieldset class="module aligned">
+				{% for field in form %}
+					<div class="form-row{% if field.errors %} errors{% endif %} field-{{ field.name }}">
```
richtier: if the field has an error, `field.css_classes` will include the value of `error_css_class` if `error_css_class` is set on the form, so you can do here instead  `form-row {{ field.css_classes }} field-{{ field.name }}`.

why are you doing `field-{{ field.name }}`?

company/tests/__init__.py
```diff
@@ -1,8 +1,28 @@
...
+import factory
+import factory.fuzzy
+
 from directory_validators.constants import choices
 from rest_framework import serializers
 
+from company.models import Company
+
+
+def ch_number():
```
richtier: the function name if not very explicit in what it does. please rename to something more helpful. 

Rule of thumb: if a two line long function needs a docstring to explain what it does, then the function name is lacking :)

company/tests/__init__.py
```diff
@@ -1,8 +1,28 @@
...
+    """Generates company house numbers for company factory"""
+    for i in range(10000000, 99999999):
+        yield str(i)
+
+
+class CompanyFactory(factory.django.DjangoModelFactory):
+
+    number = factory.Iterator(ch_number())
```
richtier: :+1:

company/admin.py
monikasulik: Tbh, I have no idea where they're copying from. This was actually talked through with Konrad and Kai. There was a file upload idea initially, but we figured that since we have no idea what format we will get and it's easy enough to do text transformation on what we do get, a comma-separated field is best.

company/admin.py
monikasulik: Yeah, I wasn't sure which way to go on this one. But this was actually what I agreed with @konradko in conversation yesterday.

company/templates/admin/company/change_list.html
monikasulik: I have no idea :D Although I'll hesitate a guess that that's short for change list.
This is a copy & paste from django/contrib/admin/templates/admin/change_list.html (the template I'm extending from) All I did was add a link to the publish view (next to the add company button), so I left all the mysterious variables there intact.

company/templates/admin/company/publish_form.html
monikasulik: This is again a copy & paste from django admin code, but I will probably be able to eliminate the use of opts all together in this now I think, so I'll have a go.

company/templates/admin/company/publish_form.html
monikasulik: I think that in this case I should always be using whatever error css class django admin does, no?

I was trying to be as close to django admin's way of styling as possible as I struggled quite a bit making it look proper and not totally misaligned. I have no idea why django adds field-fieldname to every field in admin forms, but it does. I can remove it, probably nothing bad will happen, but I kind of felt it was better to keep consistent with django's markup, no?

company/admin.py
monikasulik: Also atm we're only expecting this form to be used once or twice in its lifetime.

company/templates/admin/company/publish_form.html
richtier: ah is this copy and paste from django admin, too? OK cool.

company/admin.py
```diff
@@ -1,8 +1,50 @@
...
+    )
+
+    def clean_company_numbers(self):
+        numbers = self.cleaned_data['company_numbers'].split(',')
+        numbers = [number.strip() for number in numbers if number.strip()]
+
+        number_of_companies = Company.objects.filter(
+            number__in=numbers).count()
+        if number_of_companies != len(numbers):
+            raise forms.ValidationError(self.COMPANY_DOESNT_EXIST_MSG)
```
richtier: I'm still concerned that this message is unhelpful to the user - they will not know which company number is invalid - meaning that someone would have to spend a lot of time working out which is not valid.

something like:
```
valid_numbers = Company.objects.filter(number__in=numbers).values_list('number', flat=True)
if len(valid_numbers) != len(numbers):
    invalid_numbers = set(numbers) - set(valid_numbers)
    # include `invalid_numbers` in raise
```

PR #108, opened 2016-12-30
Allow viewing other people's case studies

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-630)

Instead of passing the sso user id in the url, we now provide it as a header. This counters the fact that now we want users to be able to see other people's case studies.

Added some permission handling to make sure that only the users attached to the case study's company can change the case study.

### DIFF
--- a/company/views.py
+++ b/company/views.py
@@ -52,7 +52,7 @@ class CompanyCaseStudyViewSet(viewsets.ModelViewSet):
         if self.request.method == 'POST':
             # on case study create, get the company using the user's sso id.
             company = generics.get_object_or_404(
-                models.Company, suppliers__sso_id=self.request.META['sso_id']
+                models.Company, suppliers__sso_id=self.request.META['sso-id']
             )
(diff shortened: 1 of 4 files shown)


company/views.py
```diff
@@ -39,31 +39,24 @@ class CompanyCaseStudyViewSet(viewsets.ModelViewSet):
...
-
-        return super().dispatch(*args, **kwargs)
-
     def get_serializer(self, *args, **kwargs):
-        if 'data' in kwargs:
-            kwargs['data']['company'] = self.company.pk
-
+        if self.request.method == 'POST':
+            # on case study create, get the company using the user's sso id.
+            company = generics.get_object_or_404(
```
monikasulik: I didn't see any tests for the 404 case, are there some? :) Also 404 test cases for GET would be good (I imagine there should be a 404 if no case study for the user with that sso id exists, right?).

konradko:
@richtier as discussed, there should be a separate API endpoint for retrieving the public profiles.

PR #110, opened 2017-01-05
Add summary to Company

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-651)

Also changed how we handle null values - we raise a db error instead of coercing to string implicitly.


company/tests/test_serializers.py
```diff
@@ -181,82 +167,6 @@ def test_company_serializer_doesnt_accept_number_over_8_chars():
 
 
 @pytest.mark.django_db
-def test_company_serializer_defaults_to_empty_string():
```
monikasulik: How come this test is being removed? Shouldn't it still be passing?

company/tests/test_serializers.py
```diff
@@ -181,82 +167,6 @@ def test_company_serializer_doesnt_accept_number_over_8_chars():
...
-    assert company.website == ''
-    assert company.description == ''
-    assert company.twitter_url == ''
-    assert company.facebook_url == ''
-    assert company.linkedin_url == ''
-    assert len(company.verification_code) == 12
-
-
-@pytest.mark.django_db
-def test_company_serializer_translates_none_to_empty_string():
```
monikasulik: ditto or if not then maybe check that the expected exception is getting raised?

company/tests/test_serializers.py
```diff
@@ -295,24 +204,6 @@ def test_company_number_serializer_validators():
 
 
 @pytest.mark.django_db
-def test_company_case_study_ensure_string(case_study_data_optional_none):
```
monikasulik: ditto

company/tests/test_signals.py
```diff
@@ -111,11 +111,9 @@ def test_marks_letter_as_sent(mock_stannp_client, settings):
...
-    for index, empty_value in enumerate(empty_values):
-        data = VALID_REQUEST_DATA.copy()
-        data['number'] = '12345{index}'.format(index=index)
-        data['contact_details'] = empty_value
-        Company.objects.create(**data)
-
-        mock_stannp_client.send_letter.assert_not_called()
+    data = VALID_REQUEST_DATA.copy()
+    data['number'] = '123456'
+    data['contact_details'] = {}
```
monikasulik: I guess this is unlikely to ever be None anyway as django's default for json seems to be 'null' anyway? (or at least that's what appears in the django admin)

supplier/tests/test_admin.py
```diff
@@ -115,22 +115,22 @@ def test_download_csv_multiple_suppliers(self):
         )
 
         row_one = (
-            ',2012-01-14 12:00:00+00:00,,,,,,{pk},False,False,,,,'
-            '2012-01-14 12:00:00+00:00,,01234568,,,,'
+            '{{}},2012-01-14 12:00:00+00:00,,,,,,{pk},False,False,,,,'
```
monikasulik: `{{}}` looks pretty strange. Is that definitely what should be happening? I would have thought that even if the default is an empty json dict type thing then it should be `{}`, no?

company/tests/test_signals.py
richtier: if `None` is explicitly passed in the database will throw an `IntegrityError`, because `null=False`.

supplier/tests/test_admin.py
richtier: http://stackoverflow.com/a/5466478/904887

richtier:
@monikasulik we now want to follow the vanilla Django behaviour:

 - if we pass in a string, a string is saved to the db
 - if we don't pass in a value then the `default` is used (usually, `''`)
 - if we pass in explicitly `None`, then try and save `None`. Throw error if `null=False`.

This is preferable over the extra boilerplate we were adding to coerce explicit `None` to `''`. Lets throw errors rather than allow `directory-ui` to give invalid data to `directory-api`.

I will remove `default=''` from `CharField` and `TextField`, good call.

company/tests/test_serializers.py
richtier: it would still pass, but it's now testing default django behaviour so not very useful.

konradko: +1 not worth testing django here

PR #112, opened 2017-01-09
Add case study csv download to admin

konradko:


richtier:
can you add a test please.

just a shallow one should do - like 5 users results in 5 rows.

Deeper tests checking the values are quite brittle, so I would avoid that, given this is not user facing.

konradko:
@richtier added tests for downloading case studies

PR #115, opened 2017-01-11
Flatten contact details

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-690)


company/migrations/0029_auto_20170111_1036.py
```diff
@@ -0,0 +1,35 @@
...
+
+from django.db import migrations
+
+
+def set_contact_detials_fields(apps, schema_editor):
+    Company = apps.get_model('company', 'Company')
+    for company in Company.objects.all():
+        details = company.contact_details
+        if details:
+            company.postal_full_name = details.get('postal_full_name') or ''
```
konradko: `details.get('postal_full_name', '')`

company/models.py
```diff
@@ -86,7 +82,62 @@ class Company(TimeStampedModel):
...
         max_length=255,
         blank=True,
-        default=''
+        default='',
+    )
+    # contact details
+    mobile_number = models.CharField(
+        blank=True,
+        default='',
+        max_length=15,
```
konradko: Do we validate that length in serializers on UI? Would not be nice to error here with 500, while getting a longer number in DB is not such a big problem

company/models.py
```diff
@@ -86,7 +82,62 @@ class Company(TimeStampedModel):
...
+    )
+    postal_full_name = models.CharField(
+        max_length=255,
+        blank=True,
+        default='',
+    )
+    address_line_1 = models.CharField(
+        max_length=200,
```
konradko: Is this the stannp limit?

company/models.py
```diff
@@ -86,7 +82,62 @@ class Company(TimeStampedModel):
...
+        blank=True,
+        default='',
+    )
+    email_full_name = models.CharField(
+        max_length=200,
+        blank=True,
+        default='',
+    )
+    # deprecated. use field level contact fields instead
+    contact_details = JSONField(
```
konradko: After we deploy this change we will need another PR that removes that field - please add a ticket

supplier/tests/test_admin.py
```diff
@@ -64,44 +50,145 @@ def test_download_csv(self):
...
+            ('company__po_box', ''),
+            ('company__postal_code', 'test_postal_code'),
+            ('company__postal_full_name', 'test_full_name'),
+            ('company__sectors', '[]'),
+            ('company__summary', ''),
+            ('company__twitter_url', ''),
+            ('company__verified_with_code', 'False'),
+            ('company__website', 'http://example.com'),
+            ('company_email', 'gargoyle@example.com'),
+            ('company_email_confirmed', 'False'),
```
konradko: another one to remove, as we no longer confirm company emails

company/helpers.py
```diff
@@ -20,6 +20,19 @@
...
+    fields = [
+        'postal_full_name',
+        'address_line_1',
+        'address_line_2',
+        'locality',
+        'country',
+        'postal_code',
+        'po_box',
+    ]
+    return any(getattr(company, field) for field in fields)
```
konradko: this is not correct - just having `country` does not mean we have the address. Before if the `contact_details` field was populated we could have assumed that it contained all details, now that there are separate fields for each part we should check that all required by stannp are set.

Please make this a method on the model too, maybe `has_valid_address`?

company/serializers.py
```diff
@@ -47,7 +47,7 @@ class CompanySerializer(serializers.ModelSerializer):
     supplier_case_studies = CompanyCaseStudySerializer(
         many=True, required=False, read_only=True
     )
-    contact_details = serializers.JSONField(required=False)
+    is_address_set = serializers.SerializerMethodField()
```
konradko: not sure what this is for - shouldn't be `is_valid_address`?

company/migrations/0029_auto_20170111_1036.py
richtier: if details can have `'postal_full_name': None`, which we want to avoid setting `company.postal_full_name` to because `null=False`.

company/models.py
richtier: I will make max_length 100.

supplier/tests/test_admin.py
richtier: I will also remove `company_email_confirmation_code` in that case

company/models.py
richtier: nope. I will set it to stannp limit of 255

company/migrations/0029_auto_20170111_1036.py
konradko: ah cool

PR #121, opened 2017-01-17
Allow only image file types for image fields

konradko:


company/serializers.py
```diff
@@ -1,10 +1,37 @@
...
+                )
+            )
+
+        return file
+
+
 class CompanyCaseStudySerializer(serializers.ModelSerializer):
 
+    image_one = AllowedFormatImageField(
+        max_length=None, allow_empty_file=False, use_url=True, required=False
```
richtier: you can pass in `validators=[validate_whitelisted_filetype]` to the field to simplify this a bit

```
def validate_whitelisted_filetype(value):
    if value.image.format.upper() not in settings.ALLOWED_IMAGE_FORMATS:
        raise serializers.ValidationError(
            "Invalid image format, allowed formats: {}".format(
                ", ".join(settings.ALLOWED_IMAGE_FORMATS)
            )
        )
```

```
 models.ImageField(
          upload_to=helpers.path_and_rename_supplier_case_study,
          blank=True,
          default='',
          validators=[validate_whitelisted_filetype]
      )
```

PR #123, opened 2017-01-18
Add csv download of buyers

monikasulik:
https://uktrade.atlassian.net/browse/ED-797

### DIFF
--- a/buyer/admin.py
+++ b/buyer/admin.py
@@ -1,4 +1,8 @@
+import csv
+import datetime
+
 from django.contrib import admin
+from django.http import HttpResponse
 
 from buyer.models import Buyer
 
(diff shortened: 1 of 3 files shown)


buyer/admin.py
```diff
@@ -7,3 +11,31 @@
 class BuyerAdmin(admin.ModelAdmin):
     search_fields = ('email', 'name', 'sector')
     readonly_fields = ('created', 'modified')
+    list_display = ('name', 'email', 'sector', 'created')
+    list_filter = ('sector', )
+    actions = ['download_csv']
+
+    def download_csv(self, request, queryset):
```
konradko: As we are using the same method on 2 other admin views (companies and case studies), maybe it would be worth to refactor it into a mixin? Only `filename` and `model` change between them (it would have to include the `excluded_fields` logic too).

PR #124, opened 2017-01-20
Update buyer model to reflect changes to subscription form

monikasulik:
https://uktrade.atlassian.net/browse/ED-819


richtier:
we now need 'comment' field too as a piece I'm doing. I will add those things to this PR.

buyer/models.py
```diff
@@ -13,8 +13,17 @@ class Buyer(TimeStampedModel):
...
-    company_name = models.CharField(max_length=255)
-    country = models.CharField(max_length=255)
+    company_name = models.CharField(
+        max_length=255,
+        default='',
+        blank=True,
+    )
+    country = models.CharField(
+        max_length=255,
+        default='',
```
monikasulik: Is there still any reason to add `default=''` now that we're not allowing null? This is default in django by default AFAIK. Also `blank=True` only effects validation and I would have thought the serializer should complain if these fields are not filled out or no?

monikasulik:
Actually, also something I should have done and didn't - probably worth updating the factory: https://github.com/uktrade/directory-api/blob/master/buyer/tests/factories.py#L12

PR #125, opened 2017-01-20
Add contact app and endpoint

konradko:
[this task](https://uktrade.atlassian.net/browse/ED-832)


contact/serializers.py
```diff
@@ -0,0 +1,53 @@
...
+            sector=validated_data['sector']
+        )
+
+    def get_or_create_buyer(self, email, name, company_name, country, sector):
+        buyer, _ = Buyer.objects.get_or_create(
+            email=email,
+            name=name,
+            company_name=company_name,
+            country=country
+        )
```
richtier: `Buyer.email` is `unique=True`, so this will result in `IntegrityError` if a buyer uses a `company_name`, `name` or `country` different to what is in the row for the specified `email`.

This will solve the problem:
```py
Buyer.objects.get_or_create(
    email=email,
    defaults={
        'name': name,
        'company_name': company_name,
        'country': country,
    }
)
```

but that would mean that the `name`, `company_name`, and `country` specified here would be ignored if the Buyer's email already exists.

richtier:
I thought there was a requirement that we do not store the messages our database?

konradko:
@richtier we do not save the content of the messages, just which buyer contacted which supplier.

PR #128, opened 2017-01-30
Remove bad images from company and case study

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-814)


company/migrations/0032_remove_bad_images.py
```diff
@@ -0,0 +1,60 @@
...
+
+def remove_non_images_from_case_study(apps, schema_editor):
+    CompanyCaseStudy = apps.get_model('company', 'CompanyCaseStudy')
+    # remove images that do not end in supported types
+    set_invalid_images_to_empty_string(CompanyCaseStudy, 'image_one')
+    set_invalid_images_to_empty_string(CompanyCaseStudy, 'image_two')
+    set_invalid_images_to_empty_string(CompanyCaseStudy, 'image_three')
+
+
+def noop(*args, **kwargs):
```
richtier: no operation :)

its used in the second argument of `RunPython`. It 'handles' the reverse migration. Without a reverse migration handler we would be unable to migrate backwards, and unable to test the migration (because the test requires migrating backwards).

company/migrations/0032_remove_bad_images.py
```diff
@@ -0,0 +1,60 @@
...
+    '.tiff',
+    '.xbm',
+]
+
+
+def set_invalid_images_to_empty_string(model_class, field_name):
+    # Case-insensitive list lookup: http://stackoverflow.com/a/21468646/904887
+    q_filter = Q()
+    for item in LEGAL_IMAGES:
+        q_filter |= Q(**{'{}__iendswith'.format(field_name): item})
```
richtier: the database does not store the file's meta data or headers - it only stores the name, so to get access to the files we would need to download them from the storage backend.

We *could* download each image to the webserver, and check the image type then delete hte image, but the complexity would be much higher than the value we get from this extra check.

PR #129, opened 2017-01-31
Management command for migrating comments to zendesk

monikasulik:
https://uktrade.atlassian.net/browse/ED-874


monikasulik:
Blocked from merging until we get sandbox details to put into heroku.

PR #135, opened 2017-02-02
Automatically publish companies that meet criteria

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-938)

It's not longer possible to manually publish companies, so I've removed code related to doing that.


company/admin.py
```diff
@@ -65,20 +17,7 @@ class CompanyAdmin(admin.ModelAdmin):
...
-            url(
-                r'^publish/$',
-                self.admin_site.admin_view(
-                    PublishByCompanyHouseNumberView.as_view()
-                ),
-                name="company_company_publish"
-            ),
-        ]
-        return additional_urls + urls
+    readonly_fields = ('created', 'modified', 'is_published')
```
monikasulik: Is is_published definitely a field that should not be possible to change? What if somebody puts up something offensive and we want to unpublish them?

company/models.py
```diff
@@ -59,7 +59,14 @@ class Company(TimeStampedModel):
         blank=True,
         null=True,
     )
-    is_published = models.BooleanField(default=False)
+    is_published = models.BooleanField(
+        default=False,
+        help_text=(
+            'Companies are automatically published based on profile '
+            'completeness - they must have description or summary, be '
+            'verified, and have en email address.'
```
monikasulik: typo: an rather than en

company/admin.py
richtier: the code in signals.py will automatically set `is_published` back to `True` if we put it to `False` becasue the criteria it checks for is still met.

If we want to 'unpublish' we can untick 'is_verified', which will make the criteria no longer met, and the code will automatically set `is_published` to `False.`

monikasulik: Ah, but then we lose the info that somebody received the letter and put the code in, but good point. I think maybe something we need to check with Alfie, but after that explanation no longer feel this is a merge stopper.

richtier: cool. I ran it by him already, but we can add an explicit field if this becomes problematic. the field can be `is_hidden` or something,

PR #136, opened 2017-02-02
[WIP] Add sectors jsonfield to buyer

monikasulik:
https://uktrade.atlassian.net/browse/ED-914


buyer/models.py
```diff
@@ -13,6 +14,7 @@ class Buyer(TimeStampedModel):
         choices=choices.COMPANY_CLASSIFICATIONS,
         max_length=255,
     )
+    sectors = JSONField(default=[])
```
richtier: can you add a comment above `sector` noting that it is not obsolete and to use `sectors` instead.

buyer/serializers.py
```diff
@@ -13,4 +14,25 @@ class Meta(object):
...
+        for attr, value in validated_data.items():
+            if attr == 'sectors':
+                sectors = json.loads(instance.sectors)
+                sectors.extend(json.loads(value))
+                instance.sectors = sectors
+            else:
+                setattr(instance, attr, value)
+        instance.save()
+
+        return instance
```
richtier: [`validate_sectors`](http://www.django-rest-framework.org/api-guide/serializers/#field-level-validation) is the method you want :)

buyer/tests/test_serializers.py
```diff
@@ -1,14 +1,17 @@
...
+from buyer.tests.factories import BuyerFactory
 
 
 @pytest.mark.django_db
-def test_buyer_deserialization():
+def test_buyer_deserialization_create():
     data = {
         'email': 'jim@example.com',
         'name': 'Jim Exampleson',
         'sector': 'AEROSPACE',
```
richtier: we no longer need to populate `sector`

buyer/views.py
```diff
@@ -1,8 +1,24 @@
...
+class BuyerCreateUpdateAPIView(CreateAPIView):
 
     serializer_class = serializers.BuyerSerializer
+
+    def post(self, request, *args, **kwargs):
+        try:
+            buyer = models.Buyer.objects.get(email=request.POST['email'])
+        except models.Buyer.DoesNotExist:
+            return self.create(request, *args, **kwargs)
+        return self.update(request, buyer)
```
richtier: why is `POST` causing an update? that is rather unrestful. I would expect a `PATCH` request to perform an update.

buyer/models.py
monikasulik: I think we'll be removing sector all together :) This is just a WIP, mainly with the intention of checking with Konrad that this is the right direction as there's still quite a bit to do.

buyer/serializers.py
monikasulik: hmmm, I'm not sure I follow? I'm not actually changing validated_data['sectors'] in any way here, what I'm changing is that the values in that var get appended to the existing ones. If you have something in mind that would be cleaner, maybe you can give me an example?

buyer/views.py
monikasulik: I totally agree. This is one of the reasons why I think @konradko should take a look at this and why I thought it would have been better to wait until we have ux/designs for a new form which probably wouldn't require the same view to do both creates and updates.

buyer/models.py
konradko: ye, after migration we will have a follow-up PR. A comment with a JIRA ticket number for that would be useful though :-)

buyer/serializers.py
konradko: I wonder when `update` is called - is it just when using PATCH? With current UI we never patch, only POST

buyer/tests/test_serializers.py
konradko: +1

buyer/views.py
konradko: I agree - currently we are capturing form submissions in the buyer model, let's keep it like that for now. This should probably be a multi-select form with PUT.

monikasulik:
I'm closing this

PR #139, opened 2017-02-03
Refactor test BuyerFactory

monikasulik:
First bit of https://uktrade.atlassian.net/browse/ED-801

### DIFF
--- a/buyer/tests/test_serializers.py
+++ b/buyer/tests/test_serializers.py
@@ -1,17 +1,12 @@
 import pytest
 
 from buyer import serializers
+from buyer.tests.factories import BuyerFactory
 
 
 @pytest.mark.django_db
 def test_buyer_deserialization():
-    data = {
-        'email': 'jim@example.com',
-        'name': 'Jim Exampleson',
-        'sector': 'AEROSPACE',
-        'company_name': 'Example corp',
-        'country': 'China',
-    }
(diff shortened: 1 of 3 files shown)


buyer/tests/test_serializers.py
```diff
@@ -1,17 +1,12 @@
...
 @pytest.mark.django_db
 def test_buyer_deserialization():
-    data = {
-        'email': 'jim@example.com',
-        'name': 'Jim Exampleson',
-        'sector': 'AEROSPACE',
-        'company_name': 'Example corp',
-        'country': 'China',
-    }
+    data = BuyerFactory.stub().__dict__
```
richtier: this test is not very transparent.  we can't see `BuyerFactory.stub().__dict__` and know what keys and values are being used.

monikasulik: Yeah, I was wondering about that and figured I'd see what people said in review ;) The thing is, I haven't found a dict method in factory boy, but it does have this stub method thing which is basically just generating an object that isn't a model, but has all the attributes that are fields on the model. Running `__dict__` on that stub object thing produces a dict that is exactly what we want however (there's no extra fields or anything like that). But yeah, it's not at all transparent. The one thing it has over hardcoding is that as the fields on the model change, this will update as well.
I can put this back as it was though.

PR #142, opened 2017-02-07
Base management command for sending notifications

monikasulik:
https://uktrade.atlassian.net/browse/ED-954


api/settings.py
```diff
@@ -300,3 +300,7 @@
     'Someone is interested in your Find a Buyer profile'
 )
 CONTACT_SUPPLIER_FROM_EMAIL = os.environ['CONTACT_SUPPLIER_FROM_EMAIL']
+
+# Notification settings
+DAILY_NOTIFICATIONS = []
+WEEKLY_NOTIFICATIONS = []
```
konradko: these should be tuples

notifications/management/commands/send_notifications.py
```diff
@@ -0,0 +1,35 @@
...
+        type_notifications_map = {
+            'daily': settings.DAILY_NOTIFICATIONS,
+            'weekly': settings.WEEKLY_NOTIFICATIONS,
+        }
+        try:
+            notify_funcs = type_notifications_map[type_option]
+        except KeyError:
+            raise CommandError(
+                "%s is not a valid notification type" % type_option)
+        return [import_string(func) for func in notify_funcs]
```
konradko: don't like `import_string` usage here. Does not seem necessary for `DAILY_NOTIFICATIONS` and `WEEKLY_NOTIFICATIONS` to be in settings either (we definitely do not want any code in the environment variables). Maybe add `daily.py` and `weekly.py` / alternatively `notifications.py`? With `run_daily` and `run_weekly` functions that can be called in `handle`

monikasulik: cool, sounds good :) Will remove completely from settings then.

PR #143, opened 2017-02-08
Model for saving info on saved notifications

monikasulik:
https://uktrade.atlassian.net/browse/ED-962


notifications/models.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+NOTIFICATION_TYPES = (
+    ('no_case_studies', 'Case studies not created'),
+    ('hasnt_logged_in', 'Not logged in after first 30 days'),
+    ('verification_code_not_given', 'Verification code not supplied'),
+)
+
+
+class SupplierNotifications(models.Model):
```
konradko: Would just call this model 'EmailNotification', as these will be sent out to both buyer and supplier. Instead of fk to a model I'd just use email address the notification was sent to. Alternatively `SupplierEmailNotification` with fk and `BuyerEmailNotification` with just email.

notifications/models.py
```diff
@@ -0,0 +1,24 @@
...
+    supplier = models.ForeignKey('user.User')
+    notification_type = models.CharField(
```
konradko: maybe `category`? Odd to add the model name to the field name

notifications/models.py
monikasulik: Hmmm, I was thinking we might want to have an easy way to track which companies notifications were sent to not just which emails. Although I guess in a slightly more complex way but we could figure that out later just by email.
I'd be happier with two tables - `SupplierEmailNotification` and `BuyerEmailNotification`. Though tbh, with current requirements being what they are we don't actually need to track the emails to buyers that we send. We send them once a week no matter what happened the previous week. Hence why I didn't include a buyer table at all.

konradko: Buyer table cannot be used - there can be multiple entries for same email there. I'd be happier with one model with email, but it's fine to have these 2 separate :)

monikasulik: well another way to go about it would be generic fk but I don't think that's a good solution... I'll go with two tables then :)

PR #144, opened 2017-02-10
Case study automated email

monikasulik:
https://uktrade.atlassian.net/browse/ED-972


notifications/notifications.py
```diff
@@ -1,6 +1,41 @@
...
+from django.utils import timezone
+
+from user.models import User as Supplier
+from notifications.models import SupplierEmailNotification
+
+
 def no_case_studies():
-    # TODO: ED-918
-    pass
+    eight_days_ago = timezone.now() - timedelta(days=8)
```
monikasulik: This should possibly be a setting with 8 for default? It might help with testing to be able to set it to 1 or 0 on dev and stage.

notifications/notifications.py
```diff
@@ -1,6 +1,41 @@
...
+        date_joined__day=eight_days_ago.day,
+    )
+    supplier_emails = suppliers.values_list('company_email', flat=True)
+    text_body = render_to_string('no_case_studies_email.txt', {})
+    html_body = render_to_string('no_case_studies_email.html', {})
+
+    message = EmailMultiAlternatives(
+        subject=settings.NO_CASE_STUDIES_SUBJECT,
+        body=text_body,
+        to=supplier_emails,
```
monikasulik: Possibly this should be bcc rather than to? I think by using to, we'd be giving visibility of emails to everybody. Another option is to send one email per person. Opinions please :)

notifications/tests/test_notifications.py
```diff
@@ -0,0 +1,117 @@
...
+    nine_days_ago = timezone.now() - timedelta(days=9)
+    SupplierFactory(date_joined=seven_days_ago)
+    supplier = SupplierFactory(date_joined=eight_days_ago)
+    SupplierFactory(date_joined=nine_days_ago)
+    mail.outbox = []  # reset after emails sent by signals
+
+    notifications.no_case_studies()
+
+    assert len(mail.outbox) == 1
+    assert mail.outbox[0].to == [supplier.company_email]
```
monikasulik: Review note to myself: Should also be asserting the subject line is correct

notifications/notifications.py
```diff
@@ -1,6 +1,41 @@
...
+    eight_days_ago = timezone.now() - timedelta(days=8)
+    suppliers = Supplier.objects.filter(
+        company__supplier_case_studies__isnull=True,
+        supplieremailnotification__isnull=True,
+        date_joined__year=eight_days_ago.year,
+        date_joined__month=eight_days_ago.month,
+        date_joined__day=eight_days_ago.day,
+    )
+    supplier_emails = suppliers.values_list('company_email', flat=True)
+    text_body = render_to_string('no_case_studies_email.txt', {})
```
monikasulik: Another review note to myself: Ask Rob if he definitely doesn't need anything in the context (such as the user's name etc.)

api/settings.py
```diff
@@ -300,3 +300,5 @@
     'Someone is interested in your Find a Buyer profile'
 )
 CONTACT_SUPPLIER_FROM_EMAIL = os.environ['CONTACT_SUPPLIER_FROM_EMAIL']
+
+NO_CASE_STUDIES_SUBJECT = "No case studies created"
```
monikasulik: Review note to myself: This should probably be an env variable with a default

notifications/notifications.py
richtier: I vote BCC

notifications/notifications.py
```diff
@@ -1,6 +1,41 @@
...
+from notifications.models import SupplierEmailNotification
+
+
 def no_case_studies():
-    # TODO: ED-918
-    pass
+    eight_days_ago = timezone.now() - timedelta(days=8)
+    suppliers = Supplier.objects.filter(
+        company__supplier_case_studies__isnull=True,
+        supplieremailnotification__isnull=True,
```
richtier: what if the supplier has a a `supplieremailnotification`, but one that is not `'no_case_studies'`? Perhaps this should do a more complex filter that filters out `supplieremailnotification` that equal `'no_case_studies'`

notifications/notifications.py
richtier: just a note - out settings has `settings.TIME_ZONE = 'UTC'`, so that simplifies things. `timezone.now()` should be the ame as `datetime.utcnow()`

richtier: good use case for making it an environment variable :)

api/templates/no_case_studies_email.txt
```diff
@@ -0,0 +1 @@
+It's been 8 days and you still have no case studies.
```
monikasulik: Yep, he does. I probably need to put the paths to the templates into those tickets once this is merged.

notifications/notifications.py
monikasulik: No harm in keeping it as `timezone.now()` as that way if someday for some reason that setting changes it should still work fine :)

notifications/notifications.py
monikasulik: Very good point. Will add a test for this as well.

notifications/notifications.py
```diff
@@ -1,6 +1,41 @@
...
+        date_joined__year=eight_days_ago.year,
+        date_joined__month=eight_days_ago.month,
+        date_joined__day=eight_days_ago.day,
```
monikasulik: Yep, daily.

api/settings.py
konradko: +1

notifications/notifications.py
konradko: +1

konradko: Time in docker containers is UTC as well, so should be fine. On the other hand, we will never want not-UTC, so would be actually good to use UTC explicitly.

notifications/notifications.py
konradko: Email addresses of other users must not be disclosed.

This probably should also be a single message to each recipient as well, it is likely that we will want to include individualised content, such as addressing by name.

notifications/notifications.py
```diff
@@ -1,6 +1,43 @@
...
+        )
+        message.attach_alternative(html_body, "text/html")
+        message.send()
+
+    notification_objs = [
+        SupplierEmailNotification(supplier=supplier,
+                                  category='no_case_studies')
+        for supplier in suppliers
+    ]
+    SupplierEmailNotification.objects.bulk_create(notification_objs)
```
konradko: I see how this is more optimal, but also means that if any email fails to send we won't save any notification.

I think having a `send_no_case_studies_notification` on Supplier / User model and just calling `supplier.send_no_case_studies_notification` here would be nice too. Not a required change though :)

monikasulik: I'm not sure I like the idea of having the method on the Supplier model personally. It means that methods on the Supplier model would be checking for and creating db objects from a completely different django app. Plus this change would be a lot of work at this point.

But the point on all of these not getting created if just one email send fails is very valid. Will change.

konradko:
@monikasulik conflicts here

PR #146, opened 2017-02-15
Add celery

konradko:


api/settings.py
```diff
@@ -300,3 +300,12 @@
     'Someone is interested in your Find a Buyer profile'
 )
 CONTACT_SUPPLIER_FROM_EMAIL = os.environ['CONTACT_SUPPLIER_FROM_EMAIL']
+
+# Celery
+if (True if (os.getenv('CELERY_ENABLED') == 'true') else False):
```
richtier: `if os.getenv('CELERY_ENABLED') == 'true':`

konradko: true that haha

PR #148, opened 2017-02-16
Add development fixtures

richtier:
# outcome of loading the fixture:

image


monikasulik:
I guess there will be something coming up for sso as well?

PR #151, opened 2017-02-17
Verification code emails

monikasulik:
https://uktrade.atlassian.net/browse/ED-977


api/settings.py
```diff
@@ -308,6 +308,19 @@
 )
 NO_CASE_STUDIES_DAYS = int(os.getenv('NO_CASE_STUDIES_DAYS', '8'))
 
+VERIFICATION_CODE_NOT_GIVEN_SUBJECT = os.getenv(
+    'VERIFICATION_CODE_NOT_GIVEN_SUBJECT',
+    'Verification code not input',
```
richtier: I guess @robfinchtrade is defining the actual subject line? Is he aware of this need?

same question for all the other content stuff in the PR

notifications/notifications.py
```diff
@@ -8,6 +8,25 @@
 from notifications import models, constants
 
 
+def _send_email_notifications(
```
richtier: starting with an underscore makes it harder to read. I assume this is meant to indicate the function is private to the module? if so you can use`__all__` to whitelist what things you want to expose from this module.

api/settings.py
robfinchtrade: Yep. I am trying to specify subject line for emails. Does sender need to specified too?

monikasulik: I don't think sender needs to be specified unless there's some good reason for it?

robfinchtrade: A good sender name simply reassures people about who it's come from and helps understand it isn't spam/may be worth opening.
eg. "Find a buyer service" might be a better name than "no-reply@find-a-buyer.export.great.gov.uk".

monikasulik: That's up to you. Quite possibly senders is something that we should look at for all emails we send rather than just specific emails in isolation.

robfinchtrade: Sure. I'll make a JIRA ticket to review it. I need to look into best practice and what the rules might be here in government.

PR #152, opened 2017-02-17
Upload logos to local file system on local environment

richtier:
I've uploaded files using `STORAGE_CLASS_NAME` as both `default` (resulting in uploading to S3) and as `local-storage` (resulting in uploading to local storage).

Both worked swimmingly.


api/settings.py
```diff
@@ -100,15 +100,17 @@
 STATIC_ROOT = os.path.join(BASE_DIR, 'staticfiles')
 if not os.path.exists(STATIC_ROOT):
     os.makedirs(STATIC_ROOT)
-
 STATIC_HOST = os.environ.get('STATIC_HOST', '')
 STATIC_URL = STATIC_HOST + '/api-static/'
 STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
+MEDIA_ROOT = os.path.join(BASE_DIR, 'media')
+MEDIA_URL = '/media/'
```
richtier: we need to set the `MEDIA_*` settings for uploading using `FileSystemStorage`

api/urls.py
```diff
@@ -107,3 +108,12 @@
         name='company-public-profile-contact-create'
     ),
 ]
+    
+if settings.STORAGE_CLASS_NAME == 'local-storage':
```
richtier: I don't like this, however, we need to make Django aware of the `/media/` root in dev env.

@monikasulik @konradko any ideas?

company/helpers.py
```diff
@@ -78,8 +78,8 @@ def __call__(self, instance, filename):
         return os.path.join(self.path, random_filename)
 
 
-path_and_rename_logos = PathAndRename(sub_path="/company_logos")
+path_and_rename_logos = PathAndRename(sub_path="company_logos")
```
richtier: @konradko when uploading to S3, `company_logos` works just as well as `/company_logos`. Was there a reason for the leading slash?

api/settings.py
monikasulik: I suggest putting in a comment about this in the code, something like `# S3 storage does not use these settings, needed only for dev local storage`. (unless I've misunderstood you?)

api/urls.py
monikasulik: I agree it is a bit icky, but I don't see how you're going to avoid something like this tbh.

PR #153, opened 2017-02-18
Email reminders for users who have not logged in for 30 days

monikasulik:
https://uktrade.atlassian.net/browse/ED-1010
DO NOT MERGE. This will need changes and a migration on sso side (last log in field).


notifications/notifications.py
```diff
@@ -48,8 +48,22 @@ def no_case_studies():
...
-    # TODO: ED-921
-    pass
+    days_ago = datetime.utcnow() - timedelta(
+        days=settings.HASNT_LOGGED_IN_DAYS)
+    suppliers = Supplier.objects.filter(
+        last_login__year=days_ago.year,
+        last_login__month=days_ago.month,
+        last_login__day=days_ago.day,
+    ).exclude(
+        supplieremailnotification__category=constants.HASNT_LOGGED_IN,
```
monikasulik: Need to check this logic with Alfie - quite possibly this email should be sent more than once.

richtier:
@monikasulik can you add a feature flag for this (and other similar automated emails) please - because it is possible we will not have the content or the template defined by the time we deploy master to prod.

monikasulik:
@richtier This will be run with a management command (which already exists). So as long as we don't schedule the management command to run daily then no emails will be sent.

user/models.py
```diff
@@ -38,6 +38,10 @@ class User(TimeStampedModel):
         _('date joined'),
         default=timezone.now,
     )
+    last_login = models.DateTimeField(
```
konradko: How will this field be set? Sessions are held on SSO. Mind that we have long sessions (lasting 30 days), so unless `last_login` is `last_active`, we can be messaging active users.

konradko: My bad, our sessions are 2 weeks

monikasulik: I put this in as a stop gap to be able to finish the brunt of the work on this (initially for some reason I was certain we already stored this info on API which we don't). The PR on sso side is up: https://github.com/uktrade/directory-sso/pull/110 I can now modify this PR appropriately. So there are two possible ways of handling this - I can save the last_login info into the db once a day from the endpoint. Or I can do this without saving to db and just pull from the endpoint when needed which is probably the simpler solution at this point.

monikasulik: Do we want to change the length of the sessions then? Maybe a question to Alfie?

requirements.txt
```diff
@@ -17,3 +17,4 @@ celery==4.0.2
 git+https://git@github.com/alphagov/notifications-python-client.git@1.4.0#egg=notifications-python-client
 git+https://github.com/uktrade/directory-validators.git@1.3.2#egg=directory_validators
 git+https://github.com/uktrade/directory-constants.git@2.0.0#egg=directory_constants
+git+https://github.com/uktrade/directory-sso-api-client.git@1.0.2#egg=directory_sso_api_client
```
monikasulik: This needs to be changed once things on sso api client are merged so tests actually start passing

PR #156, opened 2017-02-27
ED-917 verification emails

stevenburnell-dit:
Requires URLs for Zend links, which should be changed under https://uktrade.atlassian.net/browse/ED-1043

Requires replacement of hardcoded image URLs which should be changed under https://uktrade.atlassian.net/browse/ED-1044

Requires correct social media links which should be changed under https://uktrade.atlassian.net/browse/ED-1045


api/templates/verification_code_not_given_2nd_email.html
```diff
@@ -1,374 +1,19 @@
...
-						<!-- EMAIL HEADER // -->
-						<!--
-							The table "emailBody" is the email's container.
-							Its width can be set to 100% for a color band
-							that spans the width of the page.
-						-->
-						<table bgcolor="#DEDEDE" border="0" cellpadding="0" cellspacing="0" width="500" id="emailHeader">
+{% block body %}
+<p>You recently created a profile for your company on the Find a Buyer service, and we sent you a verification code by post 8 days ago.</p>
+<p>You can enter the code at <a href="{{ verfication_url }}" target="_blank">{{ verfication_url|default:"http://great.gov.uk/verify" }}</a>.</p>
```
monikasulik: I suggest not adding a default that doesn't actually exist on the site :) Better no default at all. Also, could we change the url to simply `https://find-a-buyer.export.great.gov.uk/confirm-company-address` ? Sorry, I know I said to use a context var, but it turns out the code isn't in the url (and therefore isn't dynamic) and it's much easier just to hardcode.

api/templates/verification_code_not_given_2nd_email.html
```diff
@@ -1,374 +1,19 @@
...
+<p>Your profile will not be visible to international buyers until you have verified your company’s profile.</p>
+<p>If you have not received your verification code, <a href="[LINK: zendesk]" target="_blank">let us know</a> so that we can send you a new code.</p>
```
monikasulik: I guess that's another link we need to find out. Maybe create a ticket with the missing bits, mark it as blocked and assign to Kai or Alfie?

api/templates/verification_code_not_given_2nd_email.html
```diff
@@ -1,374 +1,19 @@
...
-								<td align="center" valign="top">
-									<!-- CENTERING TABLE // -->
-									<table border="0" cellpadding="0" cellspacing="0" width="100%">
-										<tr>
-											<td align="center" valign="top">
-												<!-- FLEXIBLE CONTAINER // -->
-												<table border="0" cellpadding="10" cellspacing="0" width="500" class="flexibleContainer">
-													<tr>
-														<td valign="top" width="500" class="flexibleContainerCell">
+{% block action %}
```
monikasulik: Why an empty block?

api/templates/verification_code_not_given_2nd_email.html
```diff
@@ -1,374 +1,19 @@
...
-						</table>
-						<!-- // END -->
-
-					</td>
-				</tr>
-			</table>
-		</center>
-	</body>
-</html>
+{% block footer %}
```
monikasulik: Ditto, doesn't this use the default footer?

api/templates/verification_code_not_given_2nd_email.txt
```diff
@@ -1,3 +1,21 @@
-Dear {{ full_name }}
+{% extends "email.txt" %}
+
+{% block heading %}Find a Buyer service{% endblock %}
+{% block to %}{{ full_name|default:"User" }}{% endblock %}
+
+{% block body %}
+You recently created a profile for your company on the Find a Buyer service, and we sent you a verification code by post 8 days ago.
+
+You can enter the code at {{ verfication_url|default:"http://great.gov.uk/verify" }}.
```
monikasulik: ditto

api/templates/email.html
```diff
@@ -0,0 +1,169 @@
...
+								</td>
+					</td>
+				</tr>
+			</thead>
+			
+			<tbody>
+				<tr>
+					<td style="font-size:14px; padding:20px 20px 20px 20px;" valign="top" align="left">
+						<h1>{% block heading %}Exporting is GREAT{% endblock %}</h1>
+						<p>Dear {% block to %}User{% endblock %},</p>
```
stevenburnell-dit: {% block to %}User{% endblock %}
Means, replace this block with anything in the inheriting template also identified as a block named 'to'. 

{{ full_name }}
would be a variable probably living in the inheriting template and likely encapsulated in a block named 'to'. 

I'd agree that it could well be a lot nicer to simply use the {{ full_name }} variable directly in the parent email, but that assumes all code generating emails are using the same variable. This is something that you could answer better than I.

api/templates/email.html
```diff
@@ -0,0 +1,169 @@
...
+									<a href="http://l.email.service.trade.gov.uk/rts/go2.aspx?h=265779&amp;tp=i-H9D-Q1q-7E7-2DcaF-1c-0-3E-1C0V-G-2DcaF-nNkXj"><img alt="sm-great-tw-icon" src="http://wpm.ccmp.eu/wpm/100058/ImageLibrary/images/socialbuttons/sm-great-tw-icon.gif" width="43" height="32" style="margin-bottom:10px;"></a>
+									</td>
+								</tr>
+							</table>
+					
+							{% block footer %}
+							<p>Additional footer content here</p>
+							{% endblock %}
+							
+							<p>You can <a href="http://l.email.service.trade.gov.uk/rts/go2.aspx?h=265774&amp;tp=i-H9D-Q1q-7E7-2DcaF-1c-0-3E-1C0X-G-2DcaF-lInlg&amp;x=i-H9D-Q1q-7E7-2DcaF-1c-0-3E-1C0X-G-2DcaF-lInlg">unsubscribe</a> if you no longer wish to receive our emails.</p>
```
stevenburnell-dit: As per questions detailed in https://uktrade.atlassian.net/browse/ED-981
See question 5, footer content and response...

RESPONSE: Feedback and privacy links are not required. Keep the 'unsubscribe' link (which will require functionality or URL - probably to be supplied by Monika Sulik)

Not sure if I've created a task anywhere for that. I will check.

api/templates/email.txt
```diff
@@ -0,0 +1,31 @@
+{% load external_url from constants_tags %}
+
+Exporting is GREAT
+great.gov.uk
+
+{% block heading %}Heading{% endblock %}
+--------------------------------------
+
+Dear {% block to %}User{% endblock %},
```
stevenburnell-dit: See earlier response to same.

api/templates/email.html
monikasulik: We control the name of the variable, so personally, I don't see a problem with assuming all emails will use the same variable name for this (it actually is much better to use consistent names even for code readability's sake), but happy to hear out any arguments against this :)

api/templates/email.txt
```diff
@@ -0,0 +1,31 @@
...
+{% endblock %}
+
+--------------------------------------
+You received this email because you set up an account on great.gov.uk. 
+
+{% block footer %}
+Additional footer content here
+{% endblock %}
+
+If you no longer wish to receive our emails you can unsubscribe here: [LINK HERE]
```
stevenburnell-dit: Yes. Correct.

api/templates/email.html
monikasulik: In which case @konradko please weigh in :D Cause I thought we were just creating an unsubscribe view for buyers, not suppliers?

api/templates/verification_code_not_given_2nd_email.html
stevenburnell-dit: The .../verify URL was supplied by Alfie in response to my asking what is was. As you will remember the verfication_url was from you. Either way, I'd see the injection of actual correct functionality-based content, through whatever chosen method, to be a BE decision. I have no problem with whatever you decide but would expect a hardcoded value to raise eyebrows.

api/templates/verification_code_not_given_2nd_email.html
stevenburnell-dit: Aha! This was an earlier expectation to cater for possibilities where we might, but not always, have a CTA button, much like the original emails. Having spoken with Alfie, I'm now inclined to believe we won't have them, so consider it currently safe to remove this block.

api/templates/verification_code_not_given_2nd_email.html
monikasulik: The reason why I suggest hardcoding upon looking at the code in the repo is that the actual url is defined in a different repo than this email. So in practice I would have to hardcode it somewhere anyway. IMO much more straightforward to hardcode it in the template than in the python code cause that's the first place people will look if they want to modify it.

api/templates/verification_code_not_given_2nd_email.html
stevenburnell-dit: This was also to accommodate earlier expectation of variances across emails sharing the basic template. I cannot remember which email now, but one of them has extra footer text that says something like "you received this email because..."

The block footer is catering for the variance caused by that email, but doesn't have to be used by all.

api/templates/verification_code_not_given_2nd_email.txt
stevenburnell-dit: So, I guess this is same response as the other one.

api/templates/verification_code_not_given_2nd_email.html
monikasulik: In which case I think the default email could have just the block defined with nothing in it and only the emails that need extra text would need to define this.

api/templates/verification_code_not_given_2nd_email.html
monikasulik: Re: verify URL - AFAIK it's wrong cause I get a 404 when I try it? AFAIK we also don't control anything directly on the great.gov.uk domain so I think this must be incorrect unless I'm missing something.

monikasulik:
Right, summary of stuff to change after some discussion/trying to clarify things with people:
* Remove {% block to %} and use {{ full_name }} instead
* Remove the unsubscribe link completely. Keep in mind an unsubscribe link might be needed on certain emails that we will create in the future, just not on the ones you have in PR atm.
* Correct typo in `{{ verfication_url }}` (should be `{{ verification_url }}`)
* Instead of `[LINK: zendesk]` do `{{ zendesk_url }}`
* Add correct social media links. As per what Alfie has put on slack just now it should be:
linkedin: https://www.linkedin.com/company/department-for-international-trade
Twitter: https://twitter.com/tradegovuk
* Remove the View in browser link

stevenburnell-dit:
@monikasulik Changes made as per your suggestions in last consolidated comment. Hopefully things are now less confusing.

api/templates/verification_code_not_given_2nd_email.html
```diff
@@ -6,7 +6,7 @@
 
 {% block body %}
 <p>You recently created a profile for your company on the Find a Buyer service, and we sent you a verification code by post 8 days ago.</p>
-<p>You can enter the code at <a href="{{ verfication_url }}" target="_blank">{{ verfication_url|default:"http://great.gov.uk/verify" }}</a>.</p>
+<p>You can enter the code at <a href="{{ verification_url }}" target="_blank">{{ verfication_url|default:"http://great.gov.uk/verify" }}</a>.</p>
```
monikasulik: Link text still doesn't have the typo corrected :)

api/templates/verification_code_not_given_2nd_email.html
```diff
@@ -1,374 +1,18 @@
...
-						<!-- EMAIL HEADER // -->
-						<!--
-							The table "emailBody" is the email's container.
-							Its width can be set to 100% for a color band
-							that spans the width of the page.
-						-->
-						<table bgcolor="#DEDEDE" border="0" cellpadding="0" cellspacing="0" width="500" id="emailHeader">
+{% block body %}
+<p>You recently created a profile for your company on the Find a Buyer service, and we sent you a verification code by post 8 days ago.</p>
+<p>You can enter the code at <a href="{{ verification_url }}" target="_blank">{{ verfication_url|default:"http://great.gov.uk/verify" }}</a>.</p>
```
monikasulik: Typo still not corrected in the text of the link

PR #157, opened 2017-02-27
Add Case Study email template changes inheriting from base

stevenburnell-dit:
As per https://github.com/uktrade/directory-api/pull/156, requires work against other tickets to remove any hardcoded URLs in the parent template, and add any missing links (e.g. Zend) to this one.


api/templates/no_case_studies_email.html
```diff
@@ -1,374 +1,20 @@
...
+
+{% block heading %}Find a Buyer service{% endblock %}
+{% block to %}{{ full_name|default:"User" }}{% endblock %}
+
+{% block body %}
+<p>You recently published a profile for your company on the Find a buyer service.</p>
+<p>You can add new case studies or projects to your company’s profile at any time.</p>
+<p>Companies with case studies can be found more easily on search engines. We will also display profiles with case studies more prominently to international buyers.</p>
+<p>Log in to your Find a buyer <a href="{{ url }}" target="_blank">profile</a> to add case studies.</p>
+<p>If you have a problem using the service, <a href="[LINK: zendesk]" target="_blank">let us know</a>.</p>
```
konradko: [LINK: zendesk]? Is this work in progress?

stevenburnell-dit: As per above initial note, the Zend link is still required functionality. As per Rich's suggestion, I created a separate task, https://uktrade.atlassian.net/browse/ED-1046, against the parent story so that I could progress my current ticket without leaving it in 'in progress' blocked state.

richtier: @konradko note users will not see this template because the scheduler currently is not working.

@stevenburnell-dit I would use `to do: insert zendesk link here <ticket number>` instead of `[LINK: zendesk]` as it better explains whats going on.

PR #158, opened 2017-02-27
Add verification date

monikasulik:
https://uktrade.atlassian.net/browse/ED-1038


company/signals.py
```diff
@@ -45,3 +46,10 @@ def publish_companies_that_meet_criteria(sender, instance, *args, **kwargs):
             instance.email_address and
             instance.verified_with_code
         )
+
+
+def fill_in_verification_date(sender, instance, *args, **kwargs):
+    was_verified = sender.objects.filter(
```
richtier: why not `instance.verified_with_code`?

monikasulik: Because I need the value of verified_with_code both before and after. If I don't check what it was before then I'll end up saving the verification date every time the company is saved rather than just once.

konradko: I don't really understand this logic either. Can't you just check if instance.verified_with_code and not instance.verification_date:?

monikasulik: indeed I can - fixed

PR #159, opened 2017-02-27
Add Confirmation emails to new template inheritance version

stevenburnell-dit:


richtier:
tests are failing because 'email.txt' does not exist.

konradko:
@stevenburnell-dit @richtier it looks like there are multiple PRs for same / related thing.

PR #160, opened 2017-02-28
Add url context vars to automated emails

monikasulik:
https://uktrade.atlassian.net/browse/ED-1043


api/settings.py
```diff
@@ -337,6 +337,23 @@
 VERIFICATION_CODE_NOT_GIVEN_DAYS_2ND_EMAIL = int(os.getenv(
     'VERIFICATION_CODE_NOT_GIVEN_DAYS_2ND_EMAIL', '16'))
 
+VERIFICATION_CODE_URL = os.getenv(
+    "VERIFICATION_CODE_URL", "http://great.gov.uk/verify")
```
richtier: I suspect this will not vary with environments? If not please put the value in `constants.py`. Same comment for other (potential) constants below.

api/settings.py
```diff
@@ -337,6 +337,23 @@
...
+ZENDESK_URL = os.getenv(
+    "ZENDESK_URL",
+    "https://contact-us.export.great.gov.uk/feedback/directory/"
+)
+CASE_STUDY_URL = os.getenv(
```
richtier: can you please prefix this and `LOGIN_URL` with the service they related to (`FAB_CAS...` and `SSO_LOGIN...`)

notifications/notifications.py
```diff
@@ -17,11 +17,16 @@
...
+    notification_category, extra_context
 ):
     """Helper for sending notification emails"""
     for supplier in suppliers:
-        context = {'full_name': supplier.name}
+        context = {
+            'full_name': supplier.name,
+            'zendesk_url': settings.ZENDESK_URL,
+        }
+        context.update(extra_context)
```
richtier: pretty sure this can be:

```py
        context = {
             'full_name': supplier.name,
             'zendesk_url': settings.ZENDESK_URL,
             **extra_context,
         }
```

api/settings.py
monikasulik: Won't it? I mean when sending this email on stage, it should point to the verify url on stage rather than on live, no?

api/settings.py
monikasulik: This is literally just the variable that decides what url gets into the email and IMO it doesn't really matter which site it's on. To me what would make more sense would be to make it more consistent with the other email vars, so this would then become `NO_CASE_STUDIES_URL` and be grouped with the other vars to do with that email: https://github.com/uktrade/directory-api/blob/37ac4829e2b96b988355af7cda71eec284a17894/api/settings.py#L315-L319 Does that make sense to you?

api/settings.py
richtier: oh `http://great.gov.uk/verify` redirects to `https://find-a-buyer.export.great.gov.uk/confirm-company-address`. cool.

PR #166, opened 2017-03-07
Fix duplicate companies

richtier:
todo: add tests

### DIFF
--- a/company/tests/test_views.py
+++ b/company/tests/test_views.py
@@ -379,6 +379,14 @@ def public_profile_with_case_study():
 
 
 @pytest.fixture
+def public_profile_with_case_studies():
+    company = CompanyFactory(is_published=True)
+    CompanyCaseStudyFactory(company=company)
+    CompanyCaseStudyFactory(company=company)
+    return company
+
+
+@pytest.fixture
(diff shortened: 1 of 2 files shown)


company/tests/test_views.py
```diff
@@ -379,6 +379,14 @@ def public_profile_with_case_study():
 
 
 @pytest.fixture
+def public_profile_with_case_studies():
+    company = CompanyFactory(is_published=True)
+    CompanyCaseStudyFactory(company=company)
+    CompanyCaseStudyFactory(company=company)
```
richtier: adding this fixture to tests confirms the fix - this test fails if the new filter logic was not used.

konradko: 👍

company/views.py
```diff
@@ -32,9 +32,10 @@ class CompanyPublicProfileViewSet(viewsets.ModelViewSet):
     queryset = (
         models.Company.objects
         .filter(is_published=True)
+        .annotate(supplier_count=Count('supplier_case_studies'))
```
konradko: shouldn't this be `supplier_case_studies_count` rather than `supplier_count`?

PR #167, opened 2017-03-09
Add notification of companies published in industry of interest

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1023)
[this task](https://uktrade.atlassian.net/browse/ED-1024)
[this task](https://uktrade.atlassian.net/browse/ED-1053)


notifications/notifications.py
```diff
@@ -127,16 +150,38 @@ def verification_code_not_given():
...
-    # TODO: ED-919
-    pass
+    days = settings.NEW_COMPANIES_IN_SECTOR_FREQUENCY_DAYS
+    date_published = datetime.utcnow() - timedelta(days=days)
+    notification_category = constants.NEW_COMPANIES_IN_SECTOR
+    for industry in choices.COMPANY_CLASSIFICATIONS:
+        companies = Company.objects.filter(
+            sectors__contains=industry, date_published=date_published,
+        )
+        for buyer in Buyer.objects.filter(sector=industry):
```
richtier: this is not very efficient for database operations:
 - 36 reads from `Buyer` (there are 36 sectors)
 - 36 reads from `Supplier` (there are 36 sectors)
 - an unknowable number of writes to `AnonymousEmailNotification` every time a notification is sent.

@konradko are you concerned with this? If so I can spend some time optimising the reads and writes.

I think its ok for now, given we're not expecting a huge number of buyers io suppliers.

konradko: how about we first fetch all published on that date and then iterate through it, amending to industry lists as we go? Then we just have one read and keep the results in memory. Sth like:

```
companies_from_date_published = Company.objects.filter(date_published=date_published)
companies_per_sector = defaultdict(list)
for company in companies_from_date_published:
    companies_per_sector[company.sector].append(company)
```

Bit more memory heavy, but it is unlikely we will get massive amounts of companies for each date_published, while 1 query instead of 64 is nicer.

PR #177, opened 2017-03-20
Prevent new companies in industry spam

richtier:
This weekend I received 'new companies in indutry' email eveny day listing the exact same companies. Users would be unhappy with this occurring.

The problem happened because our definition of "new companies" is "companies published within the last week", and our anti-spam tactic was "do not send the email more than once a day".

The solution was to change the anti-spam tactic to "do not sent the email more than once a week".

### DIFF
--- a/company/tests/test_signals.py
+++ b/company/tests/test_signals.py
@@ -82,7 +82,7 @@ def test_does_not_overwrite_verification_code_if_already_set(settings):
 
 
 @pytest.mark.django_db
-@mock.patch('company.stannp.stannp_client')
+@mock.patch('company.signals.stannp_client')
 def test_does_not_send_if_letter_already_sent(mock_stannp_client, settings):
     settings.FEATURE_VERIFICATION_LETTERS_ENABLED = True
     CompanyFactory(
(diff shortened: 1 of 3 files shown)


company/tests/test_signals.py
```diff
@@ -82,7 +82,7 @@ def test_does_not_overwrite_verification_code_if_already_set(settings):
 
 
 @pytest.mark.django_db
-@mock.patch('company.stannp.stannp_client')
+@mock.patch('company.signals.stannp_client')
```
richtier: we were not mocking the correct thing - some of the tests were actually making requests to stannp. Not good.

I realised this because my internet dropped out while the tests were running from van.


This change should therefore speed up our tests a bit.

richtier: we were not mocking the correct thing - some of the tests were actually making requests to stannp. Not good.

I realised this because my internet dropped out while the tests were running from the van.

This change, therefore, should  speed up our tests a bit.

notifications/helpers.py
```diff
@@ -57,10 +58,11 @@ def get_new_companies_anonymous_subscribers():
 
     """
 
+    days = settings.NEW_COMPANIES_IN_SECTOR_FREQUENCY_DAYS
     unsubscribers = AnonymousUnsubscribe.objects.all()
     notifications_sent_today = AnonymousEmailNotification.objects.filter(
         category=constants.NEW_COMPANIES_IN_SECTOR,
-        date_sent__date=datetime.today().date()
+        date_sent__date__gte=datetime.utcnow() - timedelta(days=days)
```
konradko: just using `settings.NEW_COMPANIES_IN_SECTOR_FREQUENCY_DAYS` would be more readable in timedelta.

`notifications_sent_today` should also be renamed

PR #185, opened 2017-03-22
Fix locking tasks

konradko:


notifications/tasks.py
```diff
@@ -1,31 +1,39 @@
...
 
+def lock_acquired(lock_name):
+    """ Returns False if the lock was already set in the last 24 hours
+
+    Multiple celery beat schedulers are running at the same time, which
+    results in duplicated scheduled tasks. Cache-lock mechanism is used to
+    assure that only one task gets executed. Lock expires after 24 hours
+
+    """
+    return cache.add(lock_name, 'acquired', 86400)
```
richtier: as this is already merged it's a bit late - but we can improe thi in future by replacing the magic number `86400` with `60* 60 * 24`

PR #187, opened 2017-04-04
Add "successfully unsubscribed" notification email

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-960)


konradko:
@richtier can we get a confirmation from PM that this is wanted?

PR #188, opened 2017-04-06
Shared signature auth library

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-201)

The build will not pass until the sigauth library is created.


api/settings.py
```diff
@@ -412,3 +408,7 @@
 FAB_NOTIFICATIONS_UNSUBSCRIBE_URL = os.getenv(
     'FAB_NOTIFICATIONS_UNSUBSCRIBE_URL'
 )
+
+
+# directory signature auth
+SIGNATURE_AUTH_SECRET = os.getenv("SIGNATURE_AUTH_SECRET")
```
konradko: This is called `API_CLIENT_KEY` in other repos, your name is better, but let's have PRs for changing it in all repos if we want to change it here.

api/settings.py
```diff
@@ -128,19 +127,19 @@
...
         os.makedirs(static_dir)
 
-# Application authorisation
-UI_SECRET = os.getenv("UI_SECRET")
-
 # SECURITY WARNING: keep the secret key used in production secret!
 SECRET_KEY = os.getenv("SECRET_KEY")
 
+# Application authorisation
+UI_SECRET = os.getenv("UI_SECRET")
```
konradko: we should rename this TBH, that's legacy from the directory-form

PR #190, opened 2017-04-13
Refactor notificatons to class based

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1095)


api/settings.py
```diff
@@ -85,7 +85,7 @@
 CACHES = {
     "default": {
         "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": os.getenv('REDIS_URL'),
+        "LOCATION": os.environ['REDIS_URL'],
```
konradko: why this change? We use `getenv` everywhere in this repo as there were some issues with celery worker loading with `environ`. Not sure if still outstanding, but would have to be tested + consistently applied to all other required env vars.

makefile
```diff
@@ -65,7 +65,7 @@ DOCKER_SET_DEBUG_ENV_VARS := \
 	export DIRECTORY_API_STANNP_TEST_MODE=true; \
 	export DIRECTORY_API_CONTACT_SUPPLIER_SUBJECT=debug; \
 	export DIRECTORY_API_CONTACT_SUPPLIER_FROM_EMAIL=debug; \
-	export DIRECTORY_API_REDIS_URL=debug; \
+	export DIRECTORY_API_REDIS_URL=http://27.0.0.1:6379; \
```
konradko: ???

makefile
```diff
@@ -148,9 +148,7 @@ DEBUG_SET_ENV_VARS := \
 	export STANNP_TEST_MODE=true; \
 	export CONTACT_SUPPLIER_SUBJECT=debug; \
 	export CONTACT_SUPPLIER_FROM_EMAIL=debug; \
-	export REDIS_HOST=debug; \
-	export REDIS_PORT=debug; \
-	export REDIS_PASSWORD=debug; \
+	export REDIS_URL=http://27.0.0.1:6379; \
```
konradko: typo

notifications/email.py
```diff
@@ -0,0 +1,167 @@
...
+            body=text_body,
+            to=[self.recipient.email],
+        )
+        message.attach_alternative(html_body, "text/html")
+        message.send()
+        self.record_sent()
+
+    @property
+    def recipient(self):
+        raise NotImplementedError()
```
konradko: maybe use an abstract class instead, more fit for the purpose

api/settings.py
richtier: my reason was `debug_tesst` does not work unless `REDIS_URL` is set, meaning `REDIS_URL` is required.

but as we use `getenv` everywhere I will change it back.

makefile
richtier: I think I went too far. I changed `REDIS_URL`, and also set `DIRECTORY_API_REDIS_URL` for parity.

Note `debug_test` does not work unless `REDIS_URL` is set, meaning `REDIS_URL` is required.

notifications/email.py
```diff
@@ -0,0 +1,176 @@
...
+
+class Recipient:
+
+    def __init__(self, name, email):
+        self.email = email
+        self.name = name
+
+
+class NotificationBase(abc.ABC):
+    html_template = None
```
konradko: if these attributes are required, maybe put them in `__init__` and check they are set, e.g.:

```
def __init__(self):
    for required_attribute in self.required_attributes:
        if getattr(self, required_attribute) is None:
                raise NotImplementedError("{} is required".format(required_attribute))
```

notifications/email.py
```diff
@@ -0,0 +1,176 @@
+import abc
+
+from django.core.mail import EmailMultiAlternatives
+from django.conf import settings
+from django.template.loader import render_to_string
+
+from notifications import constants, helpers, models
+
+
+class Recipient:
```
konradko: A class just to hold some data? We have dicts and named tuples for that :P

PR #196, opened 2017-04-24
Add public supplier endpoint

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1068)

image

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -14,9 +14,10 @@ from notifications.views import (
     AnonymousUnsubscribeCreateAPIView
 )
 from supplier.views import (
-    SupplierRetrieveUpdateAPIView,
     GeckoTotalRegisteredSuppliersView,
-    UnsubscribeSupplierAPIView
+    SupplierRetrieveExternalAPIView,
+    SupplierRetrieveUpdateAPIView,
+    UnsubscribeSupplierAPIView,
 )
 from enrolment.views import EnrolmentCreateAPIView
 from buyer.views import BuyerCreateAPIView
@@ -72,7 +73,7 @@ urlpatterns = [
         name='company-case-study-detail',
     ),
     url(
-        r'supplier/(?P<sso_id>[0-9]+)/$',
+        r'^supplier/(?P<sso_id>[0-9]+)/$',
         SupplierRetrieveUpdateAPIView.as_view(),
         name='supplier'
     ),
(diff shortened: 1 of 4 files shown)


api/urls.py
```diff
@@ -43,17 +44,17 @@
         documentation
     ),
     url(
-        r'enrolment/$',
+        r'^enrolment/$',
```
richtier: without these carets, the urls could match incorrectly - i.e., before I added the carets, `public/supplier/123/` was serving the view for `supplier/123/`

supplier/serializers.py
```diff
@@ -3,6 +3,26 @@
...
+    class Meta:
+        model = Supplier
+        fields = (
+            'company_email',
+            'company_number',
+            'company_export_status',
+            'company_industries',
+            'name',
+            'sso_id',
+        )
```
richtier: I considered putting the company as a nested serializer, but that would result in:
```
{
    'company_email': '...',
    'name': '...',
    'sso_id': '...',
    'company: {
        'number': '...',
        'export_status': '...',
        'industries': '...'
    }
}
```

Notice `company_email` is not nested because it is part of the supplier. This can be confusing. Developers implementing this API would probably expect `company_email` to be within `company`.

Overall, a flatter structure is more helpful here.

konradko: we should only expose the data that is in the public profile on FAS - so `company_email` should not be there. Unless it is a public endpoint to be used only by EIG services?

richtier: it will be used only by EIG services that have the signature secret

konradko: maybe we should namespace these endpoints with `PublicProtected` then? Just in case we will ever have actual public endpoints and to imply that public does not mean without auth.

supplier/serializers.py
```diff
@@ -3,6 +3,26 @@
 from user.models import User as Supplier
 
 
+class PublicProtectedSupplierSerializer(serializers.ModelSerializer):
```
konradko: Can we rename from `PublicProtected` to `External`, forgot earlier that we use that name already

supplier/serializers.py
konradko: looking at the serializer, our Supplier / Company model is quite a mess and should really just be merged together. There is a ticket logged a long time ago for this: https://uktrade.atlassian.net/browse/ED-811, I've put it into planning.

supplier/views.py
```diff
@@ -10,6 +13,12 @@
 from user.models import User as Supplier
 
 
+class PublicProtectedSupplierRetrieveAPIView(RetrieveAPIView):
```
konradko: Please rename to `SupplierRetrieveExternalAPIView` or something similar.

api/urls.py
```diff
@@ -103,6 +104,11 @@
         name='company-public-profile-list'
     ),
     url(
+        r'^api/external/supplier/(?P<sso_id>[0-9]+)/$',
```
konradko: all endpoints are `api` endpoint in the `directory-api` service, but given this is the external proxied one, let's have a comment, or even better, put it in a separate list that we will namespace like in SSO: https://github.com/uktrade/directory-sso/blob/master/config/urls.py#L106

richtier:
to re-visit once search has been completed

PR #198, opened 2017-04-25
Slice number of companies to 5 in new companies in sector

richtier:


notifications/email.py
```diff
@@ -158,5 +158,5 @@ def get_context_data(self):
         return super().get_context_data(
             company_list_url=settings.FAS_COMPANY_LIST_URL,
             utm_params=settings.NEW_COMPANIES_IN_SECTOR_UTM,
-            companies=self.companies,
+            companies=list(self.companies)[:5],
```
konradko: a comment why we do that, probably linking to a ticket, would be nice

api/templates/new_companies_in_sector_email.txt
```diff
@@ -8,16 +8,16 @@
 You asked to be notified when new UK companies in your industry join the Find a Supplier service.
 Check out these new companies that have joined since your last visit:
 
-{% for company in companies|slice:":5" %}
```
konradko: wonder why this didn't work?

richtier: I suspect it is because it's happening in a `for` loop. The docs do not show usage with a `{% for` is supported.

konradko: was thinking same but then found some examples on SO showing this as 'working'. A mistery!

PR #203, opened 2017-05-02
Add elasticsearch client

konradko:


api/settings.py
```diff
@@ -416,3 +420,17 @@
 FAB_NOTIFICATIONS_UNSUBSCRIBE_URL = os.getenv(
     'FAB_NOTIFICATIONS_UNSUBSCRIBE_URL'
 )
+
+# Elasticsearch
+elasticsearch_client = Elasticsearch(
+    hosts=[{'host': os.getenv("ELASTICSEARCH_ENDPOINT"), 'port': 443}],
```
richtier: the host can be on he local dev env - so the port should be an env var too (the default port for locally running es is 9200).

same is true for `use_ssl`.

PR #207, opened 2017-05-04
Add search elasticsearch in search view

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1331)

### DIFF
--- a/company/views.py
+++ b/company/views.py
@@ -4,7 +4,7 @@ from rest_framework import generics, viewsets, views, status
 
 from django.db.models import Case, Count, When, Value, BooleanField
 
-from company import filters, models, pagination, serializers
+from company import filters, models, pagination, search, serializers
 
 
 class CompanyNumberValidatorAPIView(generics.GenericAPIView):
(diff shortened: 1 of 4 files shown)


docker/env.json
```diff
@@ -34,6 +34,7 @@
       "FAS_FROM_EMAIL",
       "FAB_FROM_EMAIL",
       "ELASTICSEARCH_ENDPOINT",
+      "ELASTICSEARCH_PORT",
```
konradko: this is not required, has a default - should be in optional. I'm adding these env vars on the other PR too, there will be a conflict

richtier: OK I will rebase once your branch is merged

company/views.py
```diff
@@ -140,5 +140,8 @@ def get(self, request, *args, **kwargs):
 
     @staticmethod
     def get_search_results(term, page, size):
```
konradko: Given we have ES running in docker in test, we can actually add an integration test for that

PR #209, opened 2017-05-05
Fix search by sector

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1336)


company/helpers.py
```diff
@@ -8,18 +8,27 @@
...
 
 logger = logging.getLogger(__name__)
 company_profile_url = 'https://api.companieshouse.gov.uk/company/{number}'
 
 companies_house_session = requests.Session()
 
 
+def get_sector_label(sectors_value):
+    if not sectors_value:
+        return sectors_value
```
konradko: Don't get this logic

richtier: its a copy and paste of some FAB logic. I'll simplify

company/migrations/0042_auto_20170505_1148.py
```diff
@@ -0,0 +1,24 @@
...
+from __future__ import unicode_literals
+
+from django.core import management
+from django.db import migrations
+
+def add_company_search_index_and_populate(apps, schema_editor):
+    management.call_command('rebuild_elasticsearch_index')
+
+
+def noop(*args, **kwargs):
```
konradko: is this needed?

company/search.py
```diff
@@ -29,7 +31,8 @@ class CompanyDocType(DocType):
     modified = FormattedDate(date_format='%Y-%m-%dT%H:%M:%S.%fZ')
     name = field.Text()
     number = field.Text()
-    sectors = field.Text()
+    sectors = field.Text(multi=True)
+    sectors_label = field.Text(multi=True)
```
konradko: maybe `sectors_labelled`?

company/migrations/0042_auto_20170505_1148.py
richtier: its used in the `reverse` argument of the migration. Without it django throws an exception when reverse migration happens.

we want reverse migrations to run to facilitate testing migrations.

company/search.py
richtier: I think both work equally well. Feel free to change it in a follow up PR if you're passionate about the change.

PR #210, opened 2017-05-08
Public company profile endpoint

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1339)


api/urls.py
```diff
@@ -73,6 +74,11 @@
         name='company-case-study-detail',
     ),
     url(
+        r'public/supplier/(?P<sso_id>[0-9]+)/$',
```
konradko: `external` to keep the namespace consistent

PR #217, opened 2017-05-19
Refactored common code

sdonk:


buyer/admin.py
```diff
@@ -7,6 +7,29 @@
...
+            filename=filename
+        )
+    )
+
+    fieldnames = sorted(
+        [field.name for field in model._meta.get_fields()
+         if field.name not in excluded_fields]
+    )
+
+    buyers = queryset.all().values(*fieldnames)
```
konradko: `buyers` was specific to the class, here should be more general (maybe `objects` and then `for obj in objects`)

buyer/admin.py
```diff
@@ -16,29 +39,17 @@ class BuyerAdmin(admin.ModelAdmin):
...
-        response['Content-Disposition'] = (
-            'attachment; filename="find-a-buyer_buyers_{}.csv"'.format(
-                datetime.datetime.now().strftime("%Y%m%d%H%M%S")
-            )
-        )
-
-        fieldnames = sorted(
-            [field.name for field in Buyer._meta.get_fields()
-             if field.name not in self.csv_excluded_fields]
+        return generate_csv(
```
konradko: could be little nicer with a mixin, but more explicit this way so cool

PR #220, opened 2017-05-31
Make enrolment process synchronous

sdonk:


api/settings.py
```diff
@@ -408,6 +408,7 @@
 # SSO API Client
 SSO_API_CLIENT_BASE_URL = os.getenv("SSO_API_CLIENT_BASE_URL")
 SSO_SIGNATURE_SECRET = os.getenv("SSO_SIGNATURE_SECRET")
+SQS_ENROLMENT_QUEUE_ENABLED = os.getenv('SQS_ENROLMENT_QUEUE_ENABLED', True)
```
richtier: for consistency with other feature flags, can we call this `FEATURE_SQS_ENROLMENT_QUEUE_ENABLED`

enrolment/views.py
```diff
@@ -23,12 +24,17 @@ def create(self, request, *args, **kwargs):
...
         serializer.is_valid(raise_exception=True)
 
-        enrolment.queue.EnrolmentQueue().send(
-            data=json.dumps(request.data, ensure_ascii=False)
-        )
+        if settings.FEATURE_SQS_ENROLMENT_QUEUE_ENABLED:
+            enrolment.queue.EnrolmentQueue().send(
+                data=json.dumps(request.data, ensure_ascii=False)
+            )
+            status_code = status.HTTP_202_ACCEPTED
```
richtier: nice use of status codes :)

PR #221, opened 2017-05-31
Allow setting the contact email address on company create

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1443)


enrolment/serializers.py
```diff
@@ -34,6 +34,7 @@ def create_nested_objects(self, validated_data):
                 name=data['company_name'],
                 number=data['company_number'],
                 date_of_creation=data['date_of_creation'],
+                email_address=data['contact_email_address'],
```
konradko: so it is required?

richtier: yep, and FAB is now providing this data on company create.

PR #222, opened 2017-06-02
Added case study in the csv download

sdonk:


supplier/admin.py
```diff
@@ -47,11 +48,19 @@ def download_csv(self, request, queryset):
...
         fieldnames = sorted(fieldnames)
 
-        suppliers = queryset.select_related('company').all().values(
-            *fieldnames
-        )
+        suppliers = queryset.select_related('company').all().annotate(
+            company__has_case_study=Case(
+                When(
+                    company__supplier_case_studies__isnull=False,
+                     then=Value(True)
```
richtier: funky indentation here

supplier/admin.py
```diff
@@ -47,11 +48,19 @@ def download_csv(self, request, queryset):
...
+                ),
+                default=Value(False),
```
sdonk: we probably don't need default

supplier/admin.py
sdonk: Flake8 is happy ;)

sdonk: actually it's not on circleci

supplier/admin.py
sdonk: We do! Default is empty otherwise

PR #223, opened 2017-06-02
Resend letter admin action

sdonk:


company/utils.py
```diff
@@ -0,0 +1,34 @@
+import datetime
+
+from django.conf import settings
+from django.utils import timezone
+
+from company.stannp import stannp_client
+
+
+def send_letter(company):
```
richtier: can you call this something more specific such as `send_verification_letter` please.

supplier/admin.py
```diff
@@ -71,3 +72,7 @@ def download_csv(self, request, queryset):
     download_csv.short_description = (
         "Download CSV report for selected suppliers"
     )
+
+    def resend_letter(self, request, queryset):
+        for supplier in queryset.select_related('company').all():
```
richtier: to avoid user error in django admin (selecting the wrong recipient, who may have already received a letter and input the verification code), it is worth doing `queryset.exclude(verified_with_code=True)`.

Also, is `.all()` needed at the end?

company/utils.py
sdonk: there's already a signal called like that

sdonk: @richtier suggestion? I couldn't think of anything

konradko: naming it same as the signal is fine - `send_verification_letter`

sdonk: It's used in the signal, I need to use `import as` there to avoid naming collision

supplier/tests/test_admin.py
```diff
@@ -219,3 +220,39 @@ def test_download_csv_multiple_suppliers(self):
...
+        self.client.force_login(superuser)
+
+        self.freezer = freeze_time("2012-01-14 12:00:00")
+        self.freezer.start()
+
+    def tearDown(self):
+        self.freezer.stop()
+
+    @patch('supplier.admin.send_letter')
+    def test_download_csv(self, mocked_send_letter):
```
konradko: looks like wrong name

company/utils.py
konradko: or you can just import utils

supplier/tests/test_admin.py
sdonk: ?

PR #224, opened 2017-06-05
Set FEATURE_SQS_ENROLMENT_QUEUE_ENABLED to False with string

konradko:


api/settings.py
```diff
@@ -408,10 +408,10 @@
 # SSO API Client
 SSO_API_CLIENT_BASE_URL = os.getenv("SSO_API_CLIENT_BASE_URL")
 SSO_SIGNATURE_SECRET = os.getenv("SSO_SIGNATURE_SECRET")
+
+# Synchronous profile creation
 FEATURE_SQS_ENROLMENT_QUEUE_ENABLED = os.getenv(
-    'FEATURE_SQS_ENROLMENT_QUEUE_ENABLED',
-    True
-)
+    "FEATURE_SQS_ENROLMENT_QUEUE_ENABLED") != 'false'
```
sdonk: It would be good to introduce django-environ at some point, it makes handling env vars easier and less ugly http://django-environ.readthedocs.io/en/latest/#supported-types

I'll create a ticket for it

PR #228, opened 2017-06-07
Remove async profile creation

konradko:
https://uktrade.atlassian.net/browse/ED-1440

Goodnight, sweet prince.


enrolment/views.py
```diff
@@ -18,20 +13,13 @@ class EnrolmentCreateAPIView(CreateAPIView):
     http_method_names = ("post", )
 
     def create(self, request, *args, **kwargs):
```
richtier: now this view is simply calling `serializer.save`, we can remove the `create` method, and the default behaviour of `CreateAPIView` will do the leg work for us.

supplier/signals.py
```diff
@@ -1,22 +0,0 @@
...
-
-
-def send_confirmation_email(sender, instance, created, *args, **kwargs):
-    sync_creation = settings.FEATURE_SYNCHRONOUS_PROFILE_CREATION
-    if not created or not instance.company_email or sync_creation:
-        return
-
-    context = {'url': settings.COMPANY_EMAIL_CONFIRMATION_URL}
-    text_body = render_to_string('confirmation_email.txt', context)
-    html_body = render_to_string('confirmation_email.html', context)
```
richtier: we can now delete `confirmation_email.html` and `confirmation_email.txt`.

richtier:
really nice refactor :)

enrolment/views.py
konradko: Keeping this as we do not send `data` in the request - the structure expected by the serializer is created here. Will create a separate task for this

PR #232, opened 2017-06-13
Boost position of companies in search results that have only one sector

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1477)

The solution uses [function score query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html)

image

image


company/views.py
```diff
@@ -140,8 +142,28 @@ def get(self, request, *args, **kwargs):
...
         start = (page - 1) * size
         end = start + size
-        search_object = search.CompanyDocType.search()
-        response = search_object.query('match', _all=term)[start:end].execute()
-        return response.to_dict()
+        query = search.CompanyDocType.search().query(
+            'function_score',
+            query=Q('match', _all=term),
+            functions=[
+                SF('field_value_factor', field='has_single_sector')
```
richtier: `SF` is [**score factor**](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-field-value-factor) object.

PR #233, opened 2017-06-13
Refactor enrolment view

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1631)

The purpose of the `Enrolment` model was as a persistence layer in the case of queue problems occur, so the enrollment data is not lost. We no longer have the queue, so we no longer need to model.


enrolment/serializers.py
```diff
@@ -1,73 +1,21 @@
...
+        model = Company
+        fields = [
+            'export_status',
+            'company_name',
+            'company_number',
+            'contact_email_address',
+        ]
 
 
 class SMSVerificationSerializer(serializers.Serializer):
```
konradko: Can you remove all the code related to SMS verification (please add a ticket if there's still a lot).

konradko: done

PR #241, opened 2017-06-20
Disable trusted source company code on company creation

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1677)


company/apps.py
```diff
@@ -24,3 +24,7 @@ def ready(self):
             receiver=signals.store_date_published,
             sender='company.Company'
         )
+        pre_save.connect(
```
richtier: `post_save` would be better to avoid state where the signup code is deactiveted and then the company save fails -  meaning the user can not use the code when they try to re-submit the form.

However, `post_save` does not have a handy way to check "is this the first time the company is being created?", meaning we would need to disable the codes every time the company is saved with `post_save` - meaning an extra lookup, then an extra update.

The edge case should be infrequent, and affected users can raise a ticket.

company/signals.py
```diff
@@ -32,3 +31,13 @@ def store_date_published(sender, instance, *args, **kwargs):
 def save_to_elasticsearch(sender, instance, *args, **kwargs):
     if instance.is_published:
         instance.to_doc_type().save()
+
+
+def deactivate_trusted_source_signup_code(sender, instance, *args, **kwargs):
+    # django throws "AppRegistreyNotReady" if model imported at module level
```
richtier: doing this makes me sad. any ideas @konradko @sdonk

company/signals.py
```diff
@@ -32,3 +31,13 @@ def store_date_published(sender, instance, *args, **kwargs):
...
+    from enrolment.models import TrustedSourceSignupCode
```
richtier: doing this makes me sad. any ideas @konradko @sdonk

company/signals.py
```diff
@@ -32,3 +31,13 @@ def store_date_published(sender, instance, *args, **kwargs):
...
+    if not instance.pk:
+        queryset = TrustedSourceSignupCode.objects.filter(
```
konradko: why not just update the instance?

company/apps.py
konradko: maybe it would be better to just put it in the view instead of signals then?

company/signals.py
konradko: not using signals :-)

richtier: two reasons to not use signals. I will not use signals :)

PR #242, opened 2017-06-20
Fix legacy keywords

sdonk:


company/migrations/0045_auto_20170620_1426.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('company', '0044_rebuild_elasticserach_index'),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards, reverse_code=migrations.RunPython.noop)
```
richtier: > `migrations.RunPython.noop`
did not know that exists. good to know :+1:

konradko: It is not necessary though :-)

sdonk: It is if you want your migration to be reversible otherwise, Django would happily refuse to reverse further back then this migration.

PR #247, opened 2017-06-26
Added piptools

sdonk:


requirements.ini
```diff
@@ -0,0 +1,23 @@
...
+Pillow==4.0.0
+whitenoise==3.2.2
+gunicorn==19.5.0
+psycopg2==2.6.1
+raven==5.19.0
+boto3==1.4.1
+celery==4.0.2
+elasticsearch==5.3.0
+elasticsearch-dsl==5.2.0
+requests-aws4auth==0.9
```
konradko: can you please check if this is still used? I think it was related to the removed SQS code.

richtier: it also might be related to using basic auth when talking to exops on stage.

sdonk: Tests don't fail if I remove it

PR #249, opened 2017-06-27
WIP: ED-1696 Add distributed_migration management command

sdonk:


api/management/commands/distributed_migrate.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Command(MigrateCommand):
+    """Updates database schema.
+    Manages both apps with migrations and those without."""
+
+    def handle(self, *args, **options):
+        """Execute command."""
+        with cache_lock('migrations') as result:
+            if result:
```
konradko: if the lock was not acquired the command should wait till it is available. If this is easier achieved with pglock like you mentioned, please go ahead and use it.

sdonk: This command is running all the migrations (not a single one), if the lock is acquired then we don't want to wait and run the migrations again (which won't run anyway). Right?

konradko: We do want to wait till the migrations are applied (as without that the webserver will start and fail when db without the changes will be used).

sdonk: Then pglocks achieves what you want. I'll use that

PR #250, opened 2017-06-29
Add SSO authenticator

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1736)


conftest.py
```diff
@@ -87,3 +87,11 @@ def mock_elasticsearch_company_save():
     stub = patch('company.search.CompanyDocType.save')
     yield stub.start()
     stub.stop()
+
+
+@pytest.fixture
+def requests_mocker():
+    m = requests_mock.mock()
```
sdonk: I know it's only a fixture, but I'd prefer a better name for the variable m

PR #252, opened 2017-06-29
ED-1742 run functional tests post deployment

kowalcj0:


circle.yml
```diff
@@ -18,3 +18,4 @@ deployment:
     commands:
       - make heroku_deploy_dev
       - make smoke_tests
+      - make functional_tests
```
konradko: https://github.com/uktrade/directory-api/blob/master/makefile#L226 has to be changed

PR #256, opened 2017-06-30
Role based db access urls update

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1736)

Do not merge until we've updated the api client and update FAB and SUD to pass the session id to the api client.

### DIFF
--- a/user/models.py
+++ b/user/models.py
@@ -53,6 +53,9 @@ class User(TimeStampedModel):
         ),
     )
 
+    def is_authenticated(self):
+        return True
+
     class Meta:
         ordering = ('-created', '-modified')
 
--- a/user/models.py
+++ b/user/models.py
@@ -53,9 +53,6 @@ class User(TimeStampedModel):
         ),
     )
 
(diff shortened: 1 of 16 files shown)


user/models.py
```diff
@@ -53,6 +53,9 @@ class User(TimeStampedModel):
         ),
     )
 
+    def is_authenticated(self):
+        return True
```
konradko: Code smell detected. What is this?

richtier: That's what django does in it's auth User model

in anon user:
https://github.com/django/django/blob/master/django/contrib/auth/models.py#L432

in auth user:
https://github.com/django/django/blob/master/django/contrib/auth/base_user.py#L97 (edited)

I think the reason it behaves that way is `auth.User` will only be attached to the request when the user has been authenticated. `auth.AnonUser` will only be attached to request when user has not been authenticated.

perhaps django devs should have put the `is_authenticated` on the `request` rather than on the user to avoid this issue

PR #259, opened 2017-07-07
Add preverified field to company serializer

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1795)

### DIFF
--- a/company/serializers.py
+++ b/company/serializers.py
@@ -120,6 +120,7 @@ class CompanySerializer(serializers.ModelSerializer):
             'supplier_case_studies',
             'twitter_url',
             'verified_with_code',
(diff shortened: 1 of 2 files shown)


company/serializers.py
```diff
@@ -120,6 +120,7 @@ class Meta:
             'supplier_case_studies',
             'twitter_url',
             'verified_with_code',
+            'verified_with_preverified_enrolment',
```
kowalcj0: This field is already specified in migrations https://github.com/uktrade/directory-api/blob/ae8febf856aeefd67289517a48cb76c2a171d722/company/migrations/0047_auto_20170626_1553.py#L18 so LGTM

PR #265, opened 2017-07-11
Add oauth2 authenticator to supplier view

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1799)


api/settings.py
```diff
@@ -147,7 +147,7 @@
         'django_filters.rest_framework.DjangoFilterBackend',
     ),
     'DEFAULT_AUTHENTICATION_CLASSES': (
-        'supplier.authentication.SSOAuthentication',
+        'supplier.authentication.SessionAuthenticationSSO',
```
richtier: renamed to be consistent with the new oauth2 token auhtneticator: `SSOOauth2...` looks bad - so I moved the `SSO` to the end of the class name.

requirements.ini
```diff
@@ -19,5 +19,7 @@ elasticsearch==5.3.0
 elasticsearch-dsl==5.2.0
 -e git+https://github.com/uktrade/directory-validators.git@v2.2.0#egg=directory_validators
 -e git+https://github.com/uktrade/directory-constants.git@v2.2.0#egg=directory_constants
--e git+https://github.com/uktrade/directory-sso-api-client.git@v1.3.0#egg=directory_sso_api_client
+-e git+https://github.com/uktrade/directory-sso-api-client.git@ED-1799-bearer-token-auth#egg=directory_sso_api_client
```
richtier: to update with release number when the release is ready,

makefile
```diff
@@ -67,7 +67,7 @@ DOCKER_SET_DEBUG_ENV_VARS := \
 	export DIRECTORY_API_REDIS_URL=debug; \
 	export DIRECTORY_API_STORAGE_CLASS_NAME=local-storage; \
 	export DIRECTORY_API_SSO_SIGNATURE_SECRET=api_signature_debug; \
-	export DIRECTORY_API_SSO_API_CLIENT_BASE_URL=http://sso.trade.great.dev:8004/api/v1/;\
+	export DIRECTORY_API_SSO_API_CLIENT_BASE_URL=http://sso.trade.great.dev:8004/;\
```
konradko: this is needed on FAB and SUD as well

supplier/authentication.py
```diff
@@ -38,3 +38,34 @@ def authenticate_credentials(self, session_id):
...
+        response = helpers.sso_api_client.user.get_oauth2_user_profile(
+            bearer_token
+        )
+        if not response.ok:
+            raise exceptions.AuthenticationFailed(self.message_invalid_session)
+        sso_id = response.json()['id']
+        try:
+            user = Supplier.objects.get(sso_id=sso_id)
+        except Supplier.DoesNotExist:
+            raise exceptions.AuthenticationFailed(self.message_does_not_exist)
```
konradko: ATM this endpoint on SUD returns 404 if company profile does not exist for the user

supplier/views.py
```diff
@@ -1,32 +1,26 @@
...
+from supplier import authentication, serializers, gecko
 from api.auth import GeckoBasicAuthentication
 from user.models import User as Supplier
 from notifications import notifications
 
 
-class SupplierRetrieveExternalAPIView(RetrieveAPIView):
-    lookup_field = 'sso_id'
-    queryset = Supplier.objects.all()
+class SupplierRetrieveExternalAPIView(APIView):
```
konradko: Isn't this the view SUD uses?

makefile
kowalcj0: @konradko do we need it in `directory-tests`?

konradko: @kowalcj0 if there are tests using SSO api endpoints then yes, but I think we don't have any

supplier/authentication.py
```diff
@@ -29,12 +27,37 @@ def authenticate_credentials(self, session_id):
...
         if not response.ok:
             raise exceptions.AuthenticationFailed(self.message_invalid_session)
-        sso_id = response.json()['id']
-        try:
-            user = Supplier.objects.get(sso_id=sso_id)
-        except Supplier.DoesNotExist:
-            raise exceptions.AuthenticationFailed(self.message_does_not_exist)
-        return (user, session_id)
+        sso_data = response.json()
+        sso_user = helpers.SSOUser(id=sso_data['id'], email=sso_data['email'])
```
konradko: It would be better for all the auth classes to reside in a separate module, in case we will want to use it outside of `supplier`

supplier/views.py
richtier: SUD's `FindABuyerView` talks to `SupplierRetrieveExternalAPIView` using the `sso_session_id` to authenticate the user.

SUD's `ExternalSupplierAPIView` talks to `SupplierRetrieveUpdateAPIView` using the bearer token to authenticate the user (currently that communication is broken, but this is how it will soon work when fixed). Note the FAB proxy ensures only GET is allowed.

supplier/authentication.py
richtier: yeah, now we no longer attach the `Supplier` to the request. it has not reason to stay here.

PR #267, opened 2017-07-12
Fix incorrect header encoding

richtier:


core/authentication.py
```diff
@@ -58,7 +58,7 @@ def authenticate(self, request):
             return None
         if len(auth) == 1:
             raise exceptions.AuthenticationFailed(self.message_bad_format)
-        return self.authenticate_credentials(auth[1])
+        return self.authenticate_credentials(auth[1].decode())
```
konradko: test detecting this would be nice

PR #270, opened 2017-07-14
Allow filtering by sector

richtier:
[this task[(https://uktrade.atlassian.net/browse/ED-1815)

### DIFF
--- a/company/tests/test_views.py
+++ b/company/tests/test_views.py
@@ -1,7 +1,7 @@
 import datetime
 import http
 from io import BytesIO
-from unittest.mock import patch, Mock
+from unittest.mock import call, patch, Mock
 
 from django.core.urlresolvers import reverse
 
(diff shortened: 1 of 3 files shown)


company/tests/test_views.py
```diff
@@ -851,45 +851,99 @@ def test_company_search(mock_get_search_results, api_client):
...
-            mock_search.assert_called_once_with(
-                body={
-                    'size': 5,
-                    'from': expected_start,
-                    'query': {
-                        'match': {
-                            '_all': 'bones'
-                        }
-                    },
+    with patch.object(es, 'search', return_value={}) as mock_search:
```
konradko: ES is running during unit tests - do we need to mock it?

richtier: yeah, `mock_search` is used to check the call arguments to make sure we called elasticsearch correctly.

PR #271, opened 2017-07-17
Make search term optional

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1819)


company/serializers.py
```diff
@@ -148,7 +148,7 @@ def validate_code(self, value):
 
 
 class CompanySearchSerializer(serializers.Serializer):
-    term = serializers.CharField()
+    term = serializers.CharField(required=False)
```
konradko: what if neither term nor sector will be provided? Do we validate that on the form?

kowalcj0: if so, do we have a unit test for it?

richtier: if there is no `term` or no `sector` then there is no data submitted, [meaning the view will not pass `data` into the form](https://github.com/uktrade/directory-ui-supplier/blob/master/company/views.py#L25) 

so if there is no `term` or `sector` then the form will not submit.

konradko: It should not rely on validation performed by another service. Can you add raising a validation error in case neither is provided?

richtier: it's not relying on api, if that is what you mean by "another service".

it is relying on the view not submitting the data though, so sure I will add some explicit validation

richtier: added.

PR #272, opened 2017-07-18
Add Company.has_exported_before and export preferences

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1844)


company/migrations/0052_remove_company_export_status.py
```diff
@@ -0,0 +1,19 @@
...
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('company', '0051_auto_20170718_0931'),
+    ]
+
+    operations = [
```
konradko: should happen in next release

company/tests/test_migrations.py
```diff
@@ -0,0 +1,33 @@
+import pytest
+
+from company.tests.factories import CompanyFactory
+
+
+def company_factory_factory(Company):
+    class HistoticCompanyFactory(CompanyFactory):
```
konradko: typo

company/migrations/0050_company_has_exported_before.py
```diff
@@ -0,0 +1,21 @@
...
+    dependencies = [
+        ('company', '0049_auto_20170630_1432'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='company',
+            name='has_exported_before',
+            field=models.BooleanField(default=False),
+            preserve_default=False,
```
kowalcj0: @richtier shouldn't we keep the defaults?

company/migrations/0051_auto_20170718_0931.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('company', '0050_company_has_exported_before'),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards, reverse_code=migrations.RunPython.noop)
```
kowalcj0: @richtier I'm not sure I have a in depth understanding of how migrations work, but why do we need to use `forwards` here? 
https://docs.djangoproject.com/en/1.11/howto/writing-migrations/#data-migrations-and-multiple-databases

company/migrations/0050_company_has_exported_before.py
richtier: Nope - the reason we set to `False` during migration is that immediately after this schema migration a data migration runs - which sets the value of `has_exported_before` depending on the value of `export_status`.

This field should not have a default value after the migration had run - `has_exported_before` should be specified explicitly because there is no sane default.

company/migrations/0051_auto_20170718_0931.py
richtier: this is a data migration (as opposed to a schema migration). `forwards` defines how the state of the database should be changed.

`forwards` sets the value of `has_exported_before` based on the value of `export_status`. Soon we will do a follow up change to remove `export_status`.

PR #277, opened 2017-07-27
Add management command for creating search test data

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1963)

image

We can easily change the content of `create_search_data.py` whenever manual testing requires different/more companies and case studies.


company/management/commands/create_search_data.py
```diff
@@ -0,0 +1,62 @@
+from django.core.management import BaseCommand
```
konradko: There should be a mention of 'test' in the command name or this should be just a fixture.

PR #278, opened 2017-07-27
Filter by multiple sectors

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1970)

### DIFF
--- a/company/serializers.py
+++ b/company/serializers.py
@@ -155,12 +155,22 @@ class CompanySearchSerializer(serializers.Serializer):
     term = serializers.CharField(required=False)
     page = serializers.IntegerField()
     size = serializers.IntegerField()
+    sectors = serializers.MultipleChoiceField(
+        choices=choices.COMPANY_CLASSIFICATIONS,
+        required=False,
+    )
+    # backwards compatibility. To remove once ED-945 is completed
     sector = serializers.ChoiceField(
         choices=choices.COMPANY_CLASSIFICATIONS,
         required=False,
     )
 
-    def validate(self, data):
-        if not data.get('term') and not data.get('sector'):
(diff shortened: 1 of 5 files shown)


company/serializers.py
```diff
@@ -155,12 +155,22 @@ class CompanySearchSerializer(serializers.Serializer):
...
 
-    def validate(self, data):
-        if not data.get('term') and not data.get('sector'):
+    def validate(self, attrs):
+        is_sector_present = attrs.get('sector') or attrs.get('sectors')
+        is_term_present = attrs.get('term')
+        if not (is_term_present or is_sector_present):
             raise serializers.ValidationError(self.MESSAGE_MISSING_SECTOR_TERM)
-        return data
+        # backwards compatibility. To remove once ED-945 is completed
```
konradko: can you mention on ED-945 to remove this as part of it

company/tests/test_views.py
```diff
@@ -908,15 +909,21 @@ def test_company_paginate_first_page(page_number, expected_start, api_client):
         response = api_client.get(reverse('company-search'), data=data)
 
         assert response.status_code == 200, response.content
-        mock_search.assert_called_once_with(
+        assert mock_search.call_count == 1
+        assert mock_search.call_args == call(
```
konradko: It would be nice to actually test the queries against ES instead of just checking mocks. ES is running during unit tests along with Postgres, so we should make use of it.

company/serializers.py
richtier: https://uktrade.atlassian.net/browse/ED-1971

company/tests/test_views.py
richtier: done

PR #279, opened 2017-07-27
ED-1942 Send email in a task

sdonk:


notifications/tasks.py
```diff
@@ -37,3 +39,32 @@ def verification_code_not_given():
 def new_companies_in_sector():
     if lock_acquired('new_companies_in_sector'):
         notifications.new_companies_in_sector()
+
+
+@app.task(autoretry_for=(TimeoutError, ))
+def send_email(subject,
```
kowalcj0: This would look a bit better:
```python
send_email(
    subject, text_body, html_body, recipient_email, from_email,
    category, supplier_id, anonymous):
```

sdonk: How's that better? :)

notifications/tasks.py
```diff
@@ -37,3 +39,32 @@ def verification_code_not_given():
...
+               anonymous):
+    message = EmailMultiAlternatives(
+        subject=subject,
+        body=text_body,
+        to=[recipient_email],
+        from_email=from_email,
+    )
+    message.attach_alternative(html_body, "text/html")
+    message.send()
+    if anonymous:
```
sdonk: this is the bit I don't like

notifications/tasks.py
richtier: stylistic choice. I find this preferable.
```
send_email(
    subject, text_body, html_body, recipient_email, from_email,
    category, supplier_id, anonymous
):
    # codez
    # codez
```

notifications/tasks.py
```diff
@@ -37,3 +39,45 @@ def verification_code_not_given():
...
+        subject=subject,
+        body=text_body,
+        to=[recipient_email],
+        from_email=from_email,
+    )
+    message.attach_alternative(html_body, "text/html")
+    message.send()
+
+
+@app.task(autoretry_for=(TimeoutError, ))
```
konradko: `TimeoutError` not imported?

sdonk: TimeoutError is built-in in Python 3

supplier/tests/test_views.py
```diff
@@ -107,12 +107,14 @@ def test_gecko_num_registered_supplier_view_rejects_incorrect_creds():
...
 @patch('api.signature.SignatureCheckPermission.has_permission', Mock)
-def test_unsubscribe_supplier(authed_client, authed_supplier):
+@patch('notifications.tasks.send_supplier_email')
+def test_unsubscribe_supplier(mock_task, authed_client, authed_supplier):
     response = authed_client.post(reverse('unsubscribe-supplier'))
 
     authed_supplier.refresh_from_db()
     assert response.status_code == http.client.OK
     assert authed_supplier.unsubscribed is True
+    assert mock_task.delay.called
```
richtier: I fear mock properties - because it's hard to tell the difference between its helper properties (`call_args`, `call_count`) and the properties it creates dynamically as we access them.

can you instead be more explicit with equality checks here `assert mock_task.delay.called is True` to avoid this problem

PR #282, opened 2017-07-31
Inflate companies in search results that have descriptions

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1962)
[this task](https://uktrade.atlassian.net/browse/ED-1959)


company/search.py
```diff
@@ -61,24 +63,26 @@ class Meta:
...
         linkedin_url=company.linkedin_url,
         logo=company.logo.url if company.logo else '',
         modified=company.modified,
-        sectors=company.sectors,
-        sectors_label=[helpers.get_sector_label(v) for v in company.sectors],
         name=company.name,
         number=company.number,
+        pk=str(company.pk),
+        sectors=company.sectors,
+        sectors_label=[helpers.get_sector_label(v) for v in company.sectors],
```
sdonk: Can you please rename `v` to something more meaningful? I know it's a temp var but it would help with reading the code in the future

richtier: If the variable was longer the code would span multiple lines, and I find that less readable than having it in one line, albeit with a crap variable name.

Normally I dislike single character variables or variables that are shortened words - but I am more tolerant of that with variables used extremely close to where it was defined, so I don't see it harming readability. Devs are unlikely to look at it and wonder "what is `v`?". Therefore I think using a longer variable name would harm readability.

PR #283, opened 2017-08-01
Inflate search results that have case studies

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1960)

### DIFF
--- a/company/views.py
+++ b/company/views.py
@@ -160,24 +160,49 @@ class CompanySearchAPIView(views.APIView):
         must_filters = []
         if sectors:
             for sector in sectors:
-                should_filters.append(query.Q("match", sectors=sector))
+                should_filters.append(query.Match(sectors=sector))
         if term:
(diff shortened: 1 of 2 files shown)


company/views.py
```diff
@@ -160,24 +160,49 @@ def get_search_results(term, page, size, sectors):
...
+                  'weight': 2,
+                  'filter': one_case_study + no_description,
+                }),
+                query.SF({
+                  'weight': 1,
+                  'filter': no_case_study + has_description,
+                }),
+
+            ],
+            boost_mode='sum',
```
richtier: using `sum` boost_mode will add the specified `weight` to results that match the specified criteria.

The outcome will be that (all other things being equal) companies with case studies and descriptions will be ranked higher than companies with no case studies - without the risk of poor matches being inflated too much simply because it has a description and a case study.

sdonk:
codecov seems to be stuck

PR #285, opened 2017-08-02
Add highlighting to search results

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-1956)

### DIFF
--- a/company/tests/test_views.py
+++ b/company/tests/test_views.py
@@ -450,6 +450,40 @@ def search_companies_data():
 
 
 @pytest.fixture
+def search_companies_highlighting_data():
+    index = Index('companies')
+    index.doc_type(CompanyDocType)
+    index.analyzer(analyzer('english'))
+    index.delete(ignore=404)
+    index.create()
+    CompanyFactory(
+        name='Wolf limited',
+        description=(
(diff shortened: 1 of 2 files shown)


company/tests/test_views.py
```diff
@@ -450,6 +450,40 @@ def search_companies_data():
 
 
 @pytest.fixture
+def search_companies_highlighting_data():
```
kowalcj0: @richtier where does the highlighting actually happens in this test?

company/views.py
```diff
@@ -200,9 +200,14 @@ def get_search_results(term, page, size, sectors):
                   'weight': 1,
                   'filter': no_case_study + has_description,
                 }),
-
             ],
             boost_mode='sum',
         )
+
+        search_object = search_object.highlight_options(
+            require_field_match=False,
```
kowalcj0: @richtier Dunno what were the exact requirements, but it seems to me that this should be set to `True`
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-highlighting.html#field-match

richtier: highlighting does not work if we set it to true.

PR #291, opened 2017-08-09
Add verified by companies house oauth2

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2028)


company/migrations/0055_auto_20170809_1006.py
```diff
@@ -0,0 +1,142 @@
+# -*- coding: utf-8 -*-
```
richtier: in a previous PR we added a validator to the fields in company and casestudy. This is an artifact of that change.

company/tests/test_migrations/test_0055_auto_20170809_1006.py
```diff
@@ -0,0 +1,14 @@
+import pytest
```
richtier: now we have the means to stably and quickly write unit tests for migrations we should start writing a test for each new migration.

company/signals.py
```diff
@@ -19,10 +19,7 @@ def publish_companies_that_meet_criteria(sender, instance, *args, **kwargs):
     if not instance.is_published:
         has_contact = bool(instance.email_address)
         has_synopsis = bool(instance.description or instance.summary)
-        is_verified = (
-            instance.verified_with_preverified_enrolment or
-            instance.verified_with_code
-        )
+        is_verified = instance.is_verified
```
konradko: 👍

company/tests/helpers.py
```diff
@@ -0,0 +1,8 @@
+from company.tests.factories import CompanyFactory
+
+
+def company_factory_factory(Company):
```
konradko: a comment explaining why and what this is would be nice

company/tests/test_migrations/test_0055_auto_20170809_1006.py
konradko: why would we? Can't see the point unless they're data migrations. If migrations are broken they simply won't apply and break the build in CI

company/tests/test_signals.py
```diff
@@ -118,86 +118,90 @@ def test_unknown_address_not_send_letters(mock_send_letter, settings):
...
-        ['d', 's',  'a@e.com', True,  False, True,  True],
-        ['',  's',  'a@e.com', True,  False, True,  True],
-        ['',  '',  '',         False, True,  True,  True],
-        ['',  '',  'a@e.com',  False, True,  True,  True],
-        ['d', '',  '',         False, True,  True,  True],
-        ['d', '',  'a@e.com',  False, True,  True,  True],
-        ['d', 's',  '',        False, True,  True,  True],
-        ['',  's',  '',        False, True,  True,  True],
-        ['d', 's',  'a@e.com', False, True,  True,  True],
-        ['',  's',  'a@e.com', False, True,  True,  True],
```
konradko: I'd say it's not a great test then :P

company/tests/test_migrations/test_0055_auto_20170809_1006.py
richtier: some schema migrations affect data soo - e.g., if the new field has a default value - like [test_0056_company_verified_with_companies_house_oauth2.py](https://github.com/uktrade/directory-api/pull/291/files#diff-856945f8c6175692825a6e098500992cR18).

However - this tests does nothing other than runs the migration, which is not useful. I'll remove this one.

company/tests/test_signals.py
richtier: i concur, and simplified it :)

PR #292, opened 2017-08-10
Add endpoint for validating oauth2 access token

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2035)

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -10,6 +10,7 @@ from company.views import (
     CompanyRetrieveUpdateAPIView,
     PublicCaseStudyViewSet,
     VerifyCompanyWithCodeAPIView,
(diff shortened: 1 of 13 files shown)


kowalcj0:
It'd be nice to move `https://account.companieshouse.gov.uk` to a variable and fix failing tests

richtier:
@kowalcj0 I'm using the hard coded URL in tests to make the tests super easy to read and also to avoid falsely asserting bad string concatenation code is ok because the same bad string concatenation code was used in tests.

Also, this makes grepping for the use of the domain easy, and find-and-replace will mean we can update the URL if we ever need to (I doubt we will).

so overall, I prefer hard-coding URLs in tests.

richtier:
code climate is drunk again:

image

image

PR #294, opened 2017-08-15
Include logo domain in search doc

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2063)

When running the app using local storage mode (serve images from filesystem, not s3) then the logo is saved in ES as "/media/thing.png". FAS will not know the domain to use.


company/search.py
```diff
@@ -21,28 +25,28 @@ def to_dict(self, *args, **kwargs):
 
 class CompanyDocType(DocType):
     case_study_count = field.Integer()
-    date_of_creation = FormattedDate(date_format='%Y-%m-%d')
+    date_of_creation = FormattedDate(date_format='%Y-%m-%d', index='no')
```
konradko: why `index='no'`?

api/settings.py
```diff
@@ -277,6 +277,7 @@
 }
 STORAGE_CLASS_NAME = os.getenv('STORAGE_CLASS_NAME', 'default')
 DEFAULT_FILE_STORAGE = STORAGE_CLASSES[STORAGE_CLASS_NAME]
+LOCAL_STORAGE_DOMAIN = os.getenv('LOCAL_STORAGE_DOMAIN')
```
konradko: can't this be done by using `AWS_S3_CUSTOM_DOMAIN`?

company/search.py
richtier: do not allow this field to be searchable.

before users could search based on "number of employees", or "primary key" or "logo path", so nonsensicle queries like "100", "4" or "thing.png" will return results - and indeed affect ordering.

api/settings.py
richtier: technically yes, but semantically no.

`AWS_S3_CUSTOM_DOMAIN` should point to an AWS S3 bucket url. `LOCAL_STORAGE_DOMAIN` points to the URL of the locally running dev server.

Note we need this because normally django determines the host and port using the headers on the request object. If no request object, then it's messy to get the server address. This was a problem when populating ES via a management command because there is no request object available.

company/search.py
konradko: sounds good, think we'll need to add 'rebuild_index' migration for this to take effect

api/settings.py
konradko: Adding logic for debug purposes only is a code smell really, how bad would it be if we use `AWS_S3_CUSTOM_DOMAIN` instead? Would just setting the env var work?

richtier: using `AWS_S3_` settings with local storage will hinder comprehension:

```py
if settings.STORAGE_CLASS_NAME == 'local-storage':
     return urljoin(settings.AWS_S3_CUSTOM_DOMAIN, url)
```

"when serviing files from the filesystem, files are server from the aws domain."

konradko: 👍

PR #295, opened 2017-08-16
Add legal industry option

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2064)

and remove obsolete code


company/migrations/0001_initial.py
```diff
@@ -21,7 +21,7 @@ class Migration(migrations.Migration):
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('description', models.TextField(blank=True, default='', null=True)),
                 ('employees', models.CharField(choices=[('', 'Please select'), ('1-10', '1-10'), ('11-50', '11-50'), ('51-200', '51-200'), ('201-500', '201-500'), ('501-1000', '501-1,000'), ('1001-10000', '1,001-10,000'), ('10001+', '10,001+')], max_length=20)),
-                ('export_status', models.CharField(choices=[('', ''), ('YES', 'Yes, in the last year.'), ('ONE_TWO_YEARS_AGO', 'Yes, 1-2 years ago.'), ('OVER_TWO_YEARS_AGO', 'Yes, over 2 years ago.'), ('NOT_YET', 'Not yet.'), ('NO_INTENTION', 'No, and we have no intention to.')], max_length=20, validators=[directory_validators.enrolment.export_status_intention])),
+                ('export_status', models.CharField(choices=[('', ''), ('YES', 'Yes, in the last year.'), ('ONE_TWO_YEARS_AGO', 'Yes, 1-2 years ago.'), ('OVER_TWO_YEARS_AGO', 'Yes, over 2 years ago.'), ('NOT_YET', 'Not yet.'), ('NO_INTENTION', 'No, and we have no intention to.')], max_length=20)),
```
richtier: I removed the `validators` becuase that validation function no longer exists.

It's "funny" that migrations include references to external code - that breaks the "self-contained" nature of migrations.

konradko: yeah, that's bad. For constants a way around it would be to wrap `makemigrations` so that it copies and hardcodes the historical state, it doesn't sound great though. Alternatively, we can just not put DB-related fields in constants repo, but that makes it less useful and we're unlikely to remember that. Any ideas how to solve this in a maintainable way?

konradko: Also, this change can possibly break the rollback.

richtier: not really - if we migrate backwards all that will happen is Django will think there are unstaged migrations because the field definition in `models.py` has `validators=[...]`, while the newest migration does not - resulting in Django prompting a new migration to be made.

PR #297, opened 2017-08-17
Fix notifications to only send email async

konradko:
https://uktrade.atlassian.net/browse/ED-2075

### DIFF
--- a/notifications/tests/test_notifications.py
+++ b/notifications/tests/test_notifications.py
@@ -10,7 +10,7 @@ from django.utils import timezone
 
 from buyer.tests.factories import BuyerFactory
 from company.tests.factories import CompanyFactory, CompanyCaseStudyFactory
-from notifications import constants, email, notifications
+from notifications import email, notifications
 from notifications.models import (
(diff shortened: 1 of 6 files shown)


notifications/tests/test_notifications.py
```diff
@@ -673,16 +666,17 @@ def test_doesnt_send_log_in_email_if_log_in_email_already_sent(mock_task):
...
         notifications.hasnt_logged_in()
 
+    assert mock_task.delay.call_count == 1
     call_args = mock_task.delay.call_args_list
     assert len(call_args) == 1
     assert call_args[0][1]['recipient_email'] == suppliers[0].company_email
-    # 2 in data setup
-    assert SupplierEmailNotification.objects.all().count() == 2
+
+    assert SupplierEmailNotification.objects.all().count() == 3
```
konradko: because of previous task mocking fewer notifications were created than there should be (see comments like https://github.com/uktrade/directory-api/pull/297/files#diff-5505105fdbd5bf23ba39882b7916090aL413)

PR #302, opened 2017-08-22
ED-2095 Removed not existent field

sdonk:


richtier:
would be good to add a test that would detect the bug.

PR #309, opened 2017-09-04
ED-2050 invite model endpoint

sdonk:


company/tests/test_views.py
```diff
@@ -1690,3 +1690,26 @@ def test_verify_companies_house_good_access_token(
...
+        reverse('transfer-ownership-invite-create'),
+        data=data,
+        format='json'
+    )
+
+    assert response.status_code == status.HTTP_201_CREATED
+    assert response.json() == data
+    assert OwnershipInvite.objects.filter(
+        new_owner_email='foo@bar.com'
+    ).exists()
```
richtier: for completeness, can you also check the other field values are correct

```
assert instance.requestor == 'ting'
```

api/urls.py
```diff
@@ -91,6 +91,11 @@
         name='company-case-study',
     ),
     url(
+        r'^supplier/company/change-ownership-invite/',
```
richtier: `transfer-ownership-invite` would be better as that is the name of the model

api/urls.py
```diff
@@ -91,6 +91,11 @@
...
+        TransferOwnershipInviteCreateView.as_view(),
+        name='transfer-ownership-invite-create'
```
richtier: `-create`

it's the HTTP method that is the determining factor if the invite is created (POST). Realistically, this endpoint could also allow retrieval.

could you rename it to `'transfer-ownership-invite`

sdonk: I wasn't sure whether to have `create` in the view name so I looked at other views and decided accordingly.
I'll keep `create` in the view for consistency, once the view changes its purpose we can remote it

PR #310, opened 2017-09-05
ED-2052 approve invite

sdonk:


company/serializers.py
```diff
@@ -206,3 +207,60 @@ class Meta:
...
+
+class OwershipInviteSerializer(serializers.ModelSerializer):
+    company_name = serializers.CharField(read_only=True, source='company.name')
+
+    def validate_new_owner_email(self, value):
+        if not self.partial:
+            return value
+        user = self.context['request'].user
+        if user.supplier is not None:
+            serializers.ValidationError('User has already a company')
```
richtier: this it the endpoint for a company owner to invite someone else to become the owner.

here `user` is the current owner, so it's ok that the existing owner alreayd has a company.

richtier: should be `raise ...`

company/serializers.py
```diff
@@ -206,3 +207,60 @@ class Meta:
...
             'requestor',
         )
+
+
+class OwershipInviteSerializer(serializers.ModelSerializer):
+    company_name = serializers.CharField(read_only=True, source='company.name')
+
+    def validate_new_owner_email(self, value):
+        if not self.partial:
+            return value
```
richtier: `new_owner_email` will only be set on create (POST). on `update` (PATCH), only `accepted` will be mutated. `validate_new_owner_email` would only be called when `new_owner_email` is included in `request.POST`.

company/serializers.py
```diff
@@ -206,3 +207,60 @@ class Meta:
...
+        user = self.context['request'].user
+        if user.supplier is not None:
+            serializers.ValidationError('User has already a company')
+        if value != user.company_email:
```
richtier: as above, this is the wrong place to put this validation - but when you do move it, I think the check should be case insensetive.

company/serializers.py
```diff
@@ -206,3 +207,60 @@ class Meta:
...
+            raise serializers.ValidationError(
+                'User accepting an incorrect invite'
+            )
+        return value
+
+    def validate_requestor(self, value):
```
richtier: as above, this wont fire during `update`.

company/serializers.py
```diff
@@ -206,3 +207,60 @@ class Meta:
...
+
+    def update(self, instance, validated_data):
+        if validated_data['accepted'] == True:
+            validated_data['accepted_date'] = now()
+        instance = super().update(instance, validated_data)
+        self.create_supplier(instance)
+        return instance
+
+    def create_supplier(self, instance):
+        from user.models import User as Supplier
```
richtier: why is this not imported at module level?

company/serializers.py
```diff
@@ -206,3 +207,60 @@ class Meta:
...
+        supplier = Supplier(
+            sso_id=self.context['request'].user.id,
```
richtier: `user.sso_id`

company/serializers.py
```diff
@@ -206,3 +207,60 @@ class Meta:
...
+        fields = (
+            'new_owner_email',
+            'company_name',
+            'company',
+            'requestor',
+            'uuid',
+            'accepted',
+        )
+
+    extra_kwargs = {
```
richtier: `extra_kwargs` should be in `Meta`

company/views.py
```diff
@@ -242,5 +240,8 @@ def get_search_results(term, page, size, sectors):
         return response.to_dict()
 
 
-class TransferOwnershipInviteCreateView(generics.CreateAPIView):
-    serializer_class = OwnershipInviteSerializer
+class TransferOwnershipInviteViewSet(viewsets.ModelViewSet):
+    serializer_class = serializers.OwershipInviteSerializer
```
richtier: typo. should be `OwershipInviteSerializer`

PR #321, opened 2017-09-22
ED-2205 multi user account content

sdonk:


api/settings.py
```diff
@@ -271,8 +271,14 @@
 FAB_FROM_EMAIL = os.getenv("FAB_FROM_EMAIL")
 FAB_OWNERSHIP_URL = os.getenv("FAB_OWNERSHIP_URL")
 FAB_COLLABORATOR_URL = os.getenv("FAB_COLLABORATOR_URL")
-OWNERSHIP_INVITE_SUBJECT = os.getenv("OWNERSHIP_INVITE_SUBJECT")
-COLLABORATOR_INVITE_SUBJECT = os.getenv("COLLABORATOR_INVITE_SUBJECT")
+OWNERSHIP_INVITE_SUBJECT = os.getenv(
+    'OWNERSHIP_INVITE_SUBJECT',
+    'Confirm ownership of {company_name}’s Find a buyer profile'
```
richtier: I suspect we want `'` instead of `’`.

api/templates/account_collaborator_email.html
```diff
@@ -0,0 +1,156 @@
+{% load email_image from enrolment_email %}
```
richtier: can you add a print screen of this rendered template please.

api/templates/account_ownership_transfer_email.txt
```diff
@@ -0,0 +1,25 @@
...
+The Department for International Trade
+
+--------------------------------------
+
+https://www.linkedin.com/company/department-for-international-trade
+https://twitter.com/tradegovuk
+
+HM Government
+http://great.gov.uk
+Department for International Trade, 3 Whitehall Pl, Westminster, London SW1A
```
richtier: This template is very similar to the other .txt, and the HTML ones are mostly the same. Can you create some base templates and inherit from them, please.

api/settings.py
sdonk: I did copy and paste from the content I was given. I'll check with Dan

company/signals.py
```diff
@@ -49,20 +49,20 @@ def save_case_study_change_to_elasticsearch(sender, instance, *args, **kwargs):
...
-        subject=instance.subject,
-        text_body=instance.invite_link,
-        html_body='<html><a href="{}">Click</a></html>'.format(
-            instance.invite_link
-        ),
-        recipient_email=instance.recipient_email,
-        from_email=FROM_EMAIL
-    )
+    notification = OwnershipChangeNotification(instance=instance)
+    notification.send()
```
richtier: can you call this `send_async`, so the future developer will get a hint it happens asynchronously.

sdonk:
screen shot 2017-09-22 at 16 47 48
screen shot 2017-09-22 at 16 47 23

PR #323, opened 2017-09-25
add-celery-and-worker-to-CF

joToui:


manifest.yml
```diff
@@ -1,5 +1,14 @@
 ---
 applications:
-- name: directory-api
+- name: directory-api-dev
```
konradko: How will we handle other environments, will there be a separate manifest file for each?

manifest.yml
```diff
@@ -1,5 +1,14 @@
...
   memory: 1G
   buildpack: python_buildpack
+  command : python manage.py distributed_migrate --noinput && python manage.py collectstatic --noinput && waitress-serve --port=$PORT api.wsgi:application
+- name: directory-api-worker-dev
+  memory: 1G
+  buildpack: python_buildpack
+  command : celery -A api worker -l info
+- name: directory-api-celery-dev
```
konradko: name should be like on heroku / in docker command name, e.g. `directory-api-celery-celery-beat-scheduler`

manifest.yml
```diff
@@ -1,5 +1,14 @@
 ---
 applications:
-- name: directory-api
+- name: directory-api-dev
   memory: 1G
   buildpack: python_buildpack
+  command : python manage.py distributed_migrate --noinput && python manage.py collectstatic --noinput && waitress-serve --port=$PORT api.wsgi:application
+- name: directory-api-worker-dev
```
konradko: like below, for consistency should be sth like `directory-api-celery-worker`

manifest.yml
joToui: not breaking your code please marge

manifest.yml
joToui: not breaking your code please marge

manifest.yml
joToui: not breaking your code please marge

joToui:
the change will be a temporary step for solving the worker and the celery issue 
so yes we will need new entry for every environment

konradko:
@joToui `need new entry for every environment` - I understand that then Jenkins ci-pipeline will take care of using correct entries for given environment.

manifest.yml
joToui: this manifest is only for dev and we will extend it once we will go to staging and prod

manifest.yml
joToui: this name will be remove in the next version of the CF 
it also a part of ci-pipeline file that have been set 
this have no visibility for the application the application is set to --no-route !!
this setting will change in the next update of this file

manifest.yml
joToui: this name will be remove in the next version of the CF
it also a part of ci-pipeline file that have been set
this have no visibility for the application the application is set to --no-route !!
this setting will change in the next update of this file

manifest.yml
konradko: Why not just update it now?

manifest.yml
konradko: Sounds good

manifest.yml
joToui: I have created 2 db links and 2 application in CF 
we are going to change the global CI-CD pipeline for this action and as soon as it will be working we are going to delete the environment and create staging so it is easier to focus on the problem we have and not on semantic this 2 application are not visible to any one ... 
if it is critical to you I will do it ... (is it really critical ? )

joToui: please replay ...

joToui:
ok so marge the file ...
or I can change the name ...

manifest.yml
joToui: please complete the marge than

PR #328, opened 2017-10-02
Upgrade to Django 1.11.5

sdonk:


api/urls.py
```diff
@@ -202,7 +203,7 @@
     urlpatterns += [
         url(
             r'^media/(?P<path>.*)$',
-            'django.views.static.serve',
+            django.views.static.serve,
```
sdonk: https://stackoverflow.com/questions/38744285/django-urls-error-view-must-be-a-callable-or-a-list-tuple-in-the-case-of-includ

PR #332, opened 2017-10-04
Add legal is great ITA recipient emails

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2287)


exportopportunity/helpers.py
```diff
@@ -1,11 +1,18 @@
...
 from django.conf import settings
 
 
 def get_ita_email(campaign, country):
+    constants = lead_generation
     campaign_email_map = {
-        'food-is-great': {
-            'france': settings.ITA_EMAILS_FOOD_IS_GREAT_FRANCE,
-            'singapore': settings.ITA_EMAILS_FOOD_IS_GREAT_SINGAPORE,
+        constants.FOOD_IS_GREAT: {
```
konradko: why not just keep the entire dictionary in settings?

richtier: this is easier to test.

this way we can change `settings` in the test and confirm the required value is used.

PR #342, opened 2017-10-13
Improve zero downtime es migration

richtier:


api/settings.py
```diff
@@ -455,6 +455,8 @@
 ELASTICSEARCH_CASE_STUDY_INDEX = os.getenv(
     'ELASTICSEARCH_CASE_STUDY_INDEX', 'casestudies'
 )
+ELASTICSEARCH_COMPANY_INDEX_ALIAS = ELASTICSEARCH_COMPANY_INDEX + 'alias'
```
konradko: what is the difference between `ELASTICSEARCH_COMPANY_INDEX` and `ELASTICSEARCH_COMPANY_INDEX_ALIAS`?

company/management/commands/elasticsearch_migrate.py
```diff
@@ -0,0 +1,97 @@
...
+        self.new_company_index = self.company_index_prefix + unique_id
+        self.new_case_study_index = self.case_study_index_prefix + unique_id
+        self.client = connections.get_connection()
+        super().__init__(*args, **kwargs)
+
+    def create_index(self, name, doc_type, alias):
+        index = Index(name)
+        index.doc_type(doc_type)
+        index.analyzer(analyzer('english'))
+        index.aliases(**{alias: {}})
```
konradko: Please add a comment explaining what happens here. Is this initialising an empty alias? Isn't that destructive to the current alias?

company/management/commands/elasticsearch_migrate.py
```diff
@@ -0,0 +1,97 @@
...
+        index.create()
+        return index
+
+    def get_indices(self, alias_name):
+        indices_client = IndicesClient(client=self.client)
+        try:
+            return list(indices_client.get_alias(name=alias_name).keys())
```
konradko: won't `.keys()` already return a list?

company/management/commands/populate_elasticsearch.py
```diff
@@ -1,24 +0,0 @@
...
-            .only('id')
-            .prefetch_related('supplier_case_studies')
-            .filter(is_published=True)
-        )
-        for company in companies:
-            method = tasks.save_company_to_elasticsearch
-            (method.delay if options['async'] else method)(pk=company.id)
-            for casestudy in company.supplier_case_studies.all():
-                method = tasks.save_case_study_to_elasticsearch
-                (method.delay if options['async'] else method)(pk=casestudy.id)
```
konradko: As the async option gets removed we will run into timeout issues on heroku again. Please add an option to execute the command asynchronously (no need for each company to be saved separately).

company/management/commands/elasticsearch_migrate.py
```diff
@@ -0,0 +1,97 @@
...
+from django.core import management
+from django.conf import settings
+
+from company import search
+from company import models
+
+
+class Command(management.BaseCommand):
+
+    help = 'Migrate elasticsearch indices.'
```
konradko: More description needed, please describe how the zero-downtime is achieved

company/tests/test_views.py
```diff
@@ -1686,6 +1689,7 @@ def test_company_search_results_highlight(search_companies_highlighting_data):
 
 
 @pytest.mark.django_db
+@pytest.mark.rebuild_elasticsearch
```
konradko: ?

company/utils.py
```diff
@@ -38,18 +35,7 @@ def send_verification_letter(company):
...
-        index.doc_type(doc_type)
-        index.analyzer(analyzer('english'))
-        index.delete(ignore=404)
-        index.create()
-
-    for company in CompanyModel.objects.filter(is_published=True):
-        search.company_model_to_doc_type(company).save()
-        for case_study in company.supplier_case_studies.all():
-            search.case_study_model_to_doc_type(case_study).save()
+    # this function was used by migrations, but has been superseded by
```
konradko: this is not nice, any better options?

core/management/commands/distributed_elasticsearch_migrate.py
```diff
@@ -0,0 +1,10 @@
+from company.management.commands.elasticsearch_migrate import (
+    Command as MigrateCommand
+)
+from django_pglocks import advisory_lock
+
+
+class Command(MigrateCommand):
+    def handle(self, *args, **options):
+        with advisory_lock(lock_id='es_migrations', shared=False, wait=True):
```
konradko: Won't this result in each instance running the command, just sequentially? If the lock exists they should just wait till it no longer exists and pass. In DB migration the command can be executed as it's idempotent.

docker/cmd-integration-test-webserver.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash -xe
 
 python /usr/src/app/manage.py migrate --noinput
+python /usr/src/app/manage.py elasticsearch_migrate
```
konradko: This means we're gonna rebuild the index on every deploy, even if it is not necessary. Is there no way to detect doctype changes/any other situations when a rebuild is actually needed?

company/management/commands/populate_elasticsearch.py
richtier: We now insert in bulk - so bye bye timeout issues :)

api/settings.py
richtier: `ELASTICSEARCH_COMPANY_INDEX_ALIAS` is the name we set the "live" index to, and when we insert documents or search documents we do the action against `ELASTICSEARCH_COMPANY_INDEX_ALIAS`, which elasticserach will automatically change to the "live" index that the alias points to.

`ELASTICSEARCH_COMPANY_INDEX` is the prefix for each new index. Realistically, we don't need it. I will remove and hard-code `company-` to during the `elasticsearc_migrate` - the value is only used in one place, after all.

company/management/commands/elasticsearch_migrate.py
richtier: will do.

note `alias` is a variable name with the value something like `companies-alias`, and the `{}` is a config object, which is empty but could contain filtering fields (so we can create an alias of an index that filters by `user id`, for example).

The outcome will be for a few milliseconds (until the old index is deleted) there are two indices with the same alias called `companies-alias`, so the old index will be used during searches until the old index is deleted.

company/management/commands/elasticsearch_migrate.py
richtier: in python 3, it returns `dict_keys([])`. I wanted to ensure `get_indices` always returns the same data type.

company/tests/test_views.py
richtier: much like `pytest.mark.django_db`, this command rebuilds elasticserach to prevent previous tests inserts into and deletes from ES affecting this test.

richtier: https://github.com/uktrade/directory-api/blob/master/conftest.py#L178

company/utils.py
richtier: we can update the migrations to not call this function. I thought that would be worse than this.

Then again, having obsolete functions doing noting sucks.

I'll update the migrations.

core/management/commands/distributed_elasticsearch_migrate.py
richtier: I'll read-read the `advisory_lock` docs.

docker/cmd-integration-test-webserver.sh
richtier: we can check the mapping in elasticsearch and compare it against the mapping generated by the doc type. Will investigate.

richtier: it's techcanilly possible, but sucks:

```
index_structure = list(commanies_index.get_mapping().values())[0]['mappings']['company_doc_type']

document_structure = CompanyDocType._doc_type.mapping.to_dict()
```

if the two values are not equal then a migration is required. however:

 - the use of "private" methods sucks
 - the classes were not meant to be used this way. hacky solution will probably break.
 - it sucks

The impact of creating a new index every deployment is tiny given we have low data, and now we use bulk insertion, it's not going to slow down deployment. We should revisit this if our index increases in size to the point its impactical though.

PR #343, opened 2017-10-18
ED-2386 store retrieve triage answers

sdonk:


exportreadiness/models.py
```diff
@@ -0,0 +1,16 @@
...
+from api.model_utils import TimeStampedModel
+
+
+class TriageResult(TimeStampedModel):
+    sector = models.CharField(max_length=255)
+    exported_before = models.BooleanField()
+    exporting_regular_part = models.BooleanField()
+    company_name = models.CharField(max_length=100)
+    sole_trader = models.BooleanField()
+    sso_id = models.PositiveIntegerField(unique=True)
```
richtier: this could be an FK to supplier. @konradko you dislike FKs, yes?

konradko: I love FKs when we have something to fk to :P Which is not the case here - user model is in another service

exportreadiness/serializers.py
```diff
@@ -0,0 +1,17 @@
...
+
+from .models import TriageResult
+
+
+class TriageResultSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = TriageResult
+        extra_kwargs = {
+            'sso_id': {'required': False}
```
konradko: this is defo required - without sso id we will be saving in session

konradko: also, does it have to be `extra_kwargs`? It is a field on `TriageResult` already

exportreadiness/models.py
```diff
@@ -0,0 +1,16 @@
+from django.db import models
+
+
+from api.model_utils import TimeStampedModel
+
+
+class TriageResult(TimeStampedModel):
```
konradko: Needs optional CH ID - should be the same name as in company

exportreadiness/models.py
```diff
@@ -0,0 +1,16 @@
...
+    sector = models.CharField(max_length=255)
+    exported_before = models.BooleanField()
+    exporting_regular_part = models.BooleanField()
```
konradko: Think `regular_exporter` is more accurate

exportreadiness/models.py
konradko: Also add `used_online_marketplaces_before`

exportreadiness/models.py
```diff
@@ -0,0 +1,16 @@
...
+    company_name = models.CharField(max_length=100)
```
konradko: this field is optional

exportreadiness/models.py
```diff
@@ -0,0 +1,16 @@
+from django.db import models
+
+
+from api.model_utils import TimeStampedModel
+
+
+class TriageResult(TimeStampedModel):
+    sector = models.CharField(max_length=255)
```
konradko: choices will be added in separate PR I understand

exportreadiness/models.py
konradko: Also, why max length 100? We should give max unless there is a good reason

exportreadiness/serializers.py
sdonk: It's marked as not required because we inject it in the `create()` method. If it was required the validation would fail and we would never get to create. The client is not passing the `sso_id` in the post data.

exportreadiness/models.py
sdonk: Where does it say that on the ticket?

exportreadiness/models.py
sdonk: `max_length` is required in Django `CharField` so pick a number. It wasn't specified in the ticket, so I picked what I thought it would be a sensible number,

exportreadiness/models.py
sdonk: That was the plan :)

exportreadiness/models.py
konradko: it says so it the comment above! :-)

exportreadiness/models.py
konradko: max possible is more sensible

exportreadiness/serializers.py
konradko: can't we set it before validation?

exportreadiness/tests/factories.py
```diff
@@ -0,0 +1,15 @@
+import factory
+
+from exportreadiness.models import TriageResult
+
+
+class TriageResultFactory(factory.django.DjangoModelFactory):
+
+    sector = 'Foo'
+    exported_before = True
```
konradko: feels like we should have a random choice from all valid answers here, maybe worth adding a ticket to improve it later

exportreadiness/serializers.py
sdonk: Why is it confusing? Technically speaking it's a piece of data we inject and we shouldn't need to validate, so it doesn't belong to `validate` imho. 
Also It's consistent with what we do in other serializers in the code base

exportreadiness/models.py
sdonk: Let me rephrase that, my comment was a bit rushed :)

Will CH id be needed in the future? If yes, we should add it when we need it (ie when we do the CH lookup)

exportreadiness/models.py
sdonk: I don't understand what you mean with max possible

exportreadiness/models.py
sdonk: I actually agree with @konradko here. Having a FK doesn't seem useful here

PR #345, opened 2017-10-19
Add choices to TriageResult sector

sdonk:


exportreadiness/models.py
```diff
@@ -1,17 +1,26 @@
...
     exported_before = models.BooleanField()
     regular_exporter = models.BooleanField()
     used_online_marketplace = models.BooleanField()
     company_name = models.CharField(max_length=255, null=True, blank=True)
     sole_trader = models.BooleanField()
     sso_id = models.PositiveIntegerField(unique=True)
 
+    @cached_property
+    def sector_name(self):
+        return comtrade_choices.CODES_SECTORS_DICT[self.sector]
```
richtier: how likey is it for choices to be removed from `CODES_SECTORS_DICT`? if any are removed in future then a KeyError could happen if we've stored an obsolete code in the db.

exportreadiness/serializers.py
```diff
@@ -4,11 +4,12 @@
 
 
 class TriageResultSerializer(serializers.ModelSerializer):
+    sector_name = serializers.CharField(read_only=True)
```
richtier: explicitly providing `sector_name` as a `serializers.CharField` will mean the attributes from the model's field are not used on the serializer field (e.g., max_length etc)

This problem can be avoided by using `extra_kwargs`.

sdonk: `sector_name` is not a model field

richtier: ah! I see

PR #348, opened 2017-10-20
ED-2397 Article read and task completed views

sdonk:


exportreadiness/models.py
```diff
@@ -41,3 +41,13 @@ def sector_name(self):
 
     def __str__(self):  # pragma: no cover
         return self.company_name
+
+
+class ArticleRead(TimeStampedModel):
+    article_uuid = models.UUIDField()
+    sso_id = models.PositiveIntegerField(unique=True)
```
sdonk: I think `unique=True` is incorrect. The combination of `article_uuid` and `sso_is` should be unique

PR #351, opened 2017-10-24
Add articles uuid choices to ArticleRead model

sdonk:


exportreadiness/models.py
```diff
@@ -7,6 +7,12 @@
 
 from api.model_utils import TimeStampedModel
 
+#  ('8b3f0454-5821-4635-a755-9e5a0214c594', 'ANALYSE_THE_COMPETITION')
```
konradko: what's this

exportreadiness/models.py
```diff
@@ -7,6 +7,12 @@
...
+EXREAD_ARTICLES_CHOICES = tuple(
+    [(getattr(exred_articles, item), item,) for
+     item in dir(exred_articles) if not item.startswith("__")]
```
konradko: what is this sorcery

exportreadiness/models.py
sdonk: An example of what the code below generates

exportreadiness/models.py
sdonk: Yeah, not a fan of it either. Explicit is better than implicit, I suppose

PR #356, opened 2017-12-04
Performance improvement of "article read" endpoint

richtier:
[part of this task](https://uktrade.atlassian.net/browse/ED-2822)

 😞 - not RESTful 
 😄 - allows bulk creation. Prevents ExRed needing to creating 40 "article read" in series during ExRed's request-response cycle.
 😄 - returns a list of read articles after bulk creating, so no need to POST then GET articles

I also calmed the logging of a few things to reduce noise.

do not merge

need to make single persist article work

### DIFF
--- a/exportreadiness/mixins.py
+++ b/exportreadiness/mixins.py
@@ -1,12 +1,17 @@
+from rest_framework.fields import empty
 from rest_framework.generics import get_object_or_404
 
+from django.http import QueryDict
 
-class InjectSSOIDCreateMixin:
 
-    def create(self, validated_data):
-        sso_id = self.context['request'].user.id
-        validated_data['sso_id'] = sso_id
-        return super().create(validated_data)
+class InjectSSOIDMixin:
+
+    def run_validation(self, data=empty):
+        if data:
+            if isinstance(data, QueryDict):
+                data = data.dict()
+            data['sso_id'] = self.context['request'].user.id
+        return super().run_validation(data)
 
 
 class GetObjectOr404FromSSOIdMixin:
--- a/exportreadiness/serializers.py
+++ b/exportreadiness/serializers.py
@@ -1,10 +1,11 @@
 from rest_framework import serializers
+from rest_framework.fields import empty
 
-from .mixins import InjectSSOIDCreateMixin
+from .mixins import InjectSSOIDMixin
 from . import models
 
 
-class TriageResultSerializer(InjectSSOIDCreateMixin,
+class TriageResultSerializer(InjectSSOIDMixin,
                              serializers.ModelSerializer):
     sector_name = serializers.CharField(read_only=True)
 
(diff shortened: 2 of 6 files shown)


exportreadiness/mixins.py
```diff
@@ -1,12 +1,13 @@
...
 
-class InjectSSOIDCreateMixin:
+class InjectSSOIDMixin:
 
-    def create(self, validated_data):
-        sso_id = self.context['request'].user.id
-        validated_data['sso_id'] = sso_id
-        return super().create(validated_data)
+    def run_validation(self, data=empty):
+        if data:
```
richtier: we now need the injection to happen before `validate` is called.

exportreadiness/serializers.py
```diff
@@ -15,17 +16,32 @@ class Meta:
...
+
+    def run_validation(self, data=empty):
+        if data:
+            uuids = {item['article_uuid'] for item in data}
+            duplicates = map(str, models.ArticleRead.objects.filter(
+                article_uuid__in=uuids
+            ).values_list('article_uuid', flat=True))
+            data = [
+                item for item in data if item['article_uuid'] not in duplicates
+            ]
```
richtier: this allows ExRed to send uuids that have already been read without error occurring thus meaning ExRed does not need to first retrieve a list of uuids it's already seen.

exportreadiness/serializers.py
```diff
@@ -15,17 +16,35 @@ class Meta:
...
+        if data:
+            # allows receiving articles that have already been read without
+            # creating duplicates - thus clients don't need to first retrieve
+            # articles it's already seen for performance gains: ED-2822
+            uuids = {item['article_uuid'] for item in data}
+            duplicates = map(str, models.ArticleRead.objects.filter(
+                article_uuid__in=uuids
+            ).values_list('article_uuid', flat=True))
+            data = [
+                item for item in data if item['article_uuid'] not in duplicates
```
sdonk: filter()

exportreadiness/mixins.py
```diff
@@ -1,12 +1,17 @@
...
-    def create(self, validated_data):
-        sso_id = self.context['request'].user.id
-        validated_data['sso_id'] = sso_id
-        return super().create(validated_data)
+class InjectSSOIDMixin:
+
+    def run_validation(self, data=empty):
+        if data:
+            if isinstance(data, QueryDict):
+               data = data.dict()
```
konradko: indentation is not a multiple of four - codeclimate valid complaint?!

exportreadiness/serializers.py
konradko: +1 - always sql if possible :-)

richtier: can you chaps clarify with a code sample what you mean

exportreadiness/mixins.py
richtier: eeeeh oops

exportreadiness/serializers.py
sdonk: Sorry, I should have been more clear. 
I didn't mean ORM `filter`, I meant `filter(lambda x: x['article_uuid'] in duplicates, items)`

PR #357, opened 2017-12-06
ED-2757 Add external triage result retrieve view

sdonk:


konradko:
Closing as we removed the cookie-based solution for triage result

PR #358, opened 2017-12-07
performance gains

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2880)


exportreadiness/migrations/0013_auto_20171205_1636.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
```
richtier: I missed this migration for [this change](https://github.com/uktrade/directory-api/pull/356/files#diff-7b6b3c1c615c7528456c0cbbbaf8bd2dR51)

exportreadiness/migrations/0014_auto_20171207_1646.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('exportreadiness', '0013_auto_20171205_1636'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='articleread',
+            name='sso_id',
+            field=models.PositiveIntegerField(db_index=True),
```
richtier: article reads are filtered on `sso_id` - so this should result in gainz

exportreadiness/serializers.py
```diff
@@ -36,6 +36,7 @@ class ArticleReadSerializer(InjectSSOIDMixin,
                             serializers.ModelSerializer):
 
     class Meta:
+        fields = ['article_uuid', 'sso_id']
```
richtier: this halves the field we're transporting over HTTP, so should result in gainz.

PR #361, opened 2017-12-11
Allow boolean fields to be null for triage

richtier:
[part of this task](https://uktrade.atlassian.net/browse/ED-2892)


exportreadiness/models.py
```diff
@@ -14,7 +14,7 @@ class TriageResult(TimeStampedModel):
         max_length=255
     )
     exported_before = models.BooleanField()
-    regular_exporter = models.BooleanField()
+    regular_exporter = models.NullBooleanField()
```
richtier: these questions are not always asked - so API should support receiving `None` here.

PR #365, opened 2017-12-19
Add more comprehensive healthcheck endpoints

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-2933)

## elasticsearch not working
image

## sso not working
image


api/urls.py
```diff
@@ -53,13 +50,28 @@
...
+        healthcheck.views.SingleSignOnAPIView.as_view(),
+        name='health-check-single-sign-on'
+    ),
+    url(
+        r'^healthcheck/elasticsearch/$',
+        healthcheck.views.ElasticsearchAPIView.as_view(),
+        name='health-check-elastic-search'
     ),
     url(
         r'^ping/$',
```
konradko: maybe `/healthcheck/ping/` for consistency?

healthcheck/backends.py
```diff
@@ -0,0 +1,33 @@
...
+
+from supplier.helpers import sso_api_client
+
+
+class ElasticSearchCheckBackend(BaseHealthCheckBackend):
+
+    def check_status(self):
+        connection = connections.get_connection()
+        if not connection.ping():
+            raise ServiceUnavailable('ping error')
```
konradko: Maybe add 'elasticsearch' into the error message, so that it is more readable in sentry?

healthcheck/backends.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def check_status(self):
+        try:
+            response = sso_api_client.ping()
+        except Exception as error:
+            raise ServiceUnavailable(str(error))
+        else:
+            if response.status_code != 200:
+                raise ServiceReturnedUnexpectedResult(
+                    self.message_bad_status.format(response)
```
konradko: ditto, 'SSO' in the error message would be more readable

healthcheck/views.py
```diff
@@ -0,0 +1,40 @@
...
+    def create_service_checker(self):
+        return SigngleSignOnBackend()
+
+
+class ElasticsearchAPIView(BaseHealthCheckAPIView):
+    def create_service_checker(self):
+        return ElasticSearchCheckBackend()
+
+
+class PingAPIView(APIView):
```
konradko: Ain't really ping, just `GET`

healthcheck/backends.py
richtier: it won't error in sentry - it's pingdom that will complain. it will show the url /healthcheck/elasticsearch/ failed with "ping error" - which gives enough info.

healthcheck/backends.py
richtier: the url pingdom will report is /healthcheck/single-sign-on/, so I think that identifies the service enough

healthcheck/views.py
richtier: indeed. calling it "ping" expresses the purpose of it well though.

perhaps "HelloWorldAPIView", "CanaryAPIView"?

PR #369, opened 2018-01-09
Merge migrations

sdonk:


exportreadiness/migrations/0016_merge_20180109_1608.py
```diff
@@ -0,0 +1,16 @@
...
+# Generated by Django 1.11.5 on 2018-01-09 16:08
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('exportreadiness', '0013_auto_20171129_1601'),
```
kowalcj0: @sdonk what about `0014_auto_20171207_1646.py`?

sdonk: @kowalcj0 ignore this PR, it was an untracked file in my env

kowalcj0: np

PR #380, opened 2018-01-29
[DNM] ED-3196 get user by email

kowalcj0:


api/urls.py
```diff
@@ -258,3 +258,11 @@
             {'document_root': settings.MEDIA_ROOT}
         ),
     ]
+
+if settings.ENABLE_INTERNAL_TESTING_API:
+    urlpatterns += [
+        url(
+            r'^internal/',
```
konradko: we already use name `external` for endpoints available to 3rd parties and implicit `internal` for comms between FAB<>API etc.

I'd suggest using a more readable `test-api` or something similar

docker-compose-test.yml
```diff
@@ -6,7 +6,7 @@ services:
     env_file: ./docker/.env-postgres.test
 
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:5.1.2
+    image: docker.elastic.co/elasticsearch/elasticsearch:5.3.1
```
konradko: please make sure this is what we use in dev/stage

api/settings.py
```diff
@@ -508,3 +508,6 @@
 CSV_DUMP_AUTH_TOKEN = os.environ['CSV_DUMP_AUTH_TOKEN']
 BUYERS_CSV_FILE_NAME = 'find-a-buyer-buyers.csv'
 SUPPLIERS_CSV_FILE_NAME = 'find-a-buyer-suppliers.csv'
+
+# internal api
+ENABLE_INTERNAL_TESTING_API = os.getenv("ENABLE_INTERNAL_TESTING_API", False)
```
konradko: Just `ENABLE_TEST_API` - we do not plan to have external test api

internal/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class InternalConfig(AppConfig):
```
konradko: the app name should be more explicit/descriptive - I'd suggest `testapi`

api/settings.py
konradko: also `ENABLE_TEST_API = os.getenv('ENABLE_INTERNAL_TESTING_API', 'false') == 'true')` - with the code in place this would evaulate to True if `ENABLE_INTERNAL_TESTING_API == 'false'`

PR #387, opened 2018-02-09
ED-3077 delete company by CH ID

kowalcj0:
This [ticket](https://uktrade.atlassian.net/browse/ED-3077)

### DIFF
--- a/testapi/views.py
+++ b/testapi/views.py
@@ -1,5 +1,9 @@
 from django.http import Http404
-from rest_framework.generics import get_object_or_404, RetrieveAPIView
+from rest_framework.generics import (
+    get_object_or_404,
+    DestroyAPIView,
+    RetrieveAPIView
+)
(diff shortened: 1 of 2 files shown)


testapi/views.py
```diff
@@ -9,23 +12,31 @@
...
             'company_email': company.email_address,
             'is_verification_letter_sent': company.is_verification_letter_sent
         }
         return Response(response_data)
+
+    def delete(self, request, ch_id):
+        try:
+            company = get_object_or_404(Company, number=ch_id)
+            company.delete()
+        except Http404:
```
richtier: no need to manually try...except...raise - that's what `get_object_or_404` does. this can be:

```
get_object_or_404(Company, number=ch_id).delete()
Response(status=204)
```

kowalcj0: indeed, much simpler

PR #391, opened 2018-02-27
ED-3534 Send to sentry if stannp response is not 200

sdonk:


company/stannp.py
```diff
@@ -38,10 +39,13 @@ def send_letter(self, template, recipient):
         data['template'] = template
         data['test'] = self.test_mode
 
-        return self.post(
+        response = self.post(
             'https://dash.stannp.com/api/v1/letters/create',
             data,
         )
+        if response.status_code != requests.codes.ok:
```
richtier: are we stannp will return non-200 on error?

PR #397, opened 2018-03-13
ED-3702 custom page 500

eadpearce:
https://uktrade.atlassian.net/browse/ED-3702


exportreadiness/migrations/0016_auto_20180313_1103.py
```diff
@@ -0,0 +1,30 @@
...
...Ships, boats and floating structures'), ('HS90', 'Optical, photographic, cinematographic, measuring, checking, medical or surgical instruments and apparatus; parts and accessories'), ('HS91', 'Clocks and watches and parts thereof'), ('HS92', 'Musical instruments; parts and accessories of such articles'), ('HS94', 'Furniture; bedding, mattresses, mattress supports, cushions and similar stuffed furnishings; lamps and lighting fittings, n.e.c.; illuminated signs, illuminated name-plates and the like; prefabricated buildings'), ('HS95', 'Toys, games and sports requisites; parts and accessories thereof'), ('HS96', 'Miscellaneous manufactured articles'), ('HS97', "Works of art; collectors' pieces and antiques"), ('HS99', 'Commodities not specified according to kind'), ('EB1', 'Transportation'), ('EB2', 'Travel'), ('EB3', 'Communications services'), ('EB4', 'Construction services'), ('EB5', 'Insurance services'), ('EB6', 'Financial services'), ('EB7', 'Computer and information services'), ('EB8', 'Royalties and license fees'), ('EB9', 'Other business services'), ('EB10', 'Personal, cultural, and recreational services'), ('EB11', 'Government services, n.i.e.')], max_length=255, null=True),
```
kowalcj0: @eadpearce I'm not sure that we have to provide all of the `choices` for that field.
To my mind those values are imported [here](https://github.com/uktrade/directory-api/blob/0707a378a438daf8f43a0b61897ff2c7f7b7ba00/exportreadiness/models.py#L40
) from [directory-constants](https://github.com/uktrade/directory-constants/blob/master/directory_constants/constants/exred_sector_names.py#L3)

eadpearce: yeah I'm not sure. that's just what was autogenerated when I made the migration so I'm wary of messing with it

richtier: Django does this because migration files should be self-contained.

If migrations relied on import that is liable to change then the migration might give different outcomes for different people, which is undesirable for migrations.

PR #398, opened 2018-03-14
Improve user experience of bulk create pre-verified enrolment

richtier:
[this task](https://uktrade.atlassian.net/browse/ED-3693)

### DIFF
--- a/enrolment/admin.py
+++ b/enrolment/admin.py
@@ -1,7 +1,10 @@
+import csv
+
 from django.conf.urls import url
 from django.contrib import admin
 from django.core.urlresolvers import reverse_lazy
-from django.views.generic import FormView
+from django.http import HttpResponse
+from django.views.generic import FormView, View
 
 from enrolment.models import PreVerifiedEnrolment
 from enrolment import forms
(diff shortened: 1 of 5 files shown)


enrolment/admin.py
```diff
@@ -20,6 +23,20 @@ def get_form_kwargs(self):
...
+
+    def get(self, request):
+        response = HttpResponse(content_type='text/csv')
+        response['Content-Disposition'] = 'attachment; filename="template.csv"'
+        writer = csv.writer(response)
+        writer.writerow(['Company number', 'Email'])
+        writer.writerow([
+            '90000001', 'comany@exmaple.com',
+            'This is an example company. Delete this row.'
+        ])
```
sdonk: Wondering if it's worth refactoring this https://github.com/uktrade/directory-api/blob/master/api/utils.py#L6 to take an iterable

enrolment/forms.py
```diff
@@ -21,7 +21,12 @@ class Meta:
 
 class GeneratePreVerifiedCompanies(forms.Form):
     generated_for = forms.CharField(max_length=1000)
-    csv_file = forms.FileField()
+    csv_file = forms.FileField(
+        help_text=(
+            '<a href="/admin/enrolment/preverifiedenrolment/example-template/"'
```
sdonk: reverse?

richtier: race condition: the reverse (or reverse_lazy) occurs at module level, before [this](https://github.com/uktrade/directory-api/blob/170e50b04f17fa2ae432c35db8f055b8dd59c4be/enrolment/admin.py#L34) is executed, so that url does not exist when we want to reverse it.

PR #414, opened 2018-05-23
WIP: Stub endpoint activity stream

michalc:


activitystream/views.py
```diff
@@ -0,0 +1,38 @@
+from django.conf import settings
+from hawkrest import HawkAuthentication
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.viewsets import ViewSet
+
+
+class ActivityStreamHawkAuthentication(HawkAuthentication):
+    def hawk_credentials_lookup(self, id):
+        if id != settings.ACTIVITY_STREAM_ACCESS_KEY_ID:
```
richtier: this is overriding the python builtin `id`, can you use a variable such as `access_key_id` please.

activitystream/views.py
```diff
@@ -0,0 +1,38 @@
+from django.conf import settings
+from hawkrest import HawkAuthentication
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.viewsets import ViewSet
+
+
+class ActivityStreamHawkAuthentication(HawkAuthentication):
```
richtier: hawk auth looks good :+1:

In case youd didn't see it, we have a permission class you could utlize if you did not need to know the authenticated user on `request.user`. in `api.signature` we have:

```
class SignatureCheckPermission(permissions.SignatureCheckPermissionBase):
    secret = settings.SIGNATURE_SECRET

```

you could similarly do:

```
class ActivityStreamSignatureCheckPermission(permissions.SignatureCheckPermissionBase):
    secret = settings.ACTIVITY_STREAM_ACCESS_KEY_ID
```

and use `ActivityStreamSignatureCheckPermission` on your view

activitystream/views.py
michalc: Changed!

activitystream/views.py
```diff
@@ -0,0 +1,117 @@
...
+class ActivityStreamAuthentication(BaseAuthentication):
+
+    # This is returned as the WWW-Authenticate header when
+    # AuthenticationFailed is raised. DRF also requires this
+    # to send a 401 (as opposed to 403)
+    def authenticate_header(self, request):
+        return 'Hawk'
+
+    def authenticate(self, request):
+        whitelist = settings.ACTIVITY_STREAM_IP_WHITELIST.split(',')
```
richtier: to be consistent with other env var settings, can you please move the type conversion in `settings.py`

activitystream/views.py
```diff
@@ -0,0 +1,117 @@
...
+
+        if 'HTTP_X_FORWARDED_FOR' not in request.META:
+            log.warning(
+                'Failed authentication: no X-Forwarded-For header passed'
+            )
+            raise AuthenticationFailed(
```
richtier: this looks like on local dev env the auth will always fail?

activitystream/views.py
```diff
@@ -0,0 +1,117 @@
...
+                lookup_credentials,
+                request.META['HTTP_AUTHORIZATION'],
+                request.build_absolute_uri(),
+                request.method,
+                content=request.body,
+                content_type=request.content_type,
+                seen_nonce=seen_nonce,
+            )
+        except HawkFail as e:
+            log.warning('Failed authentication %s', e)
```
richtier: `format` is predominantly used for string formatting in the codebase. can you use `format` here too for consistency, please

makefile
```diff
@@ -181,6 +184,9 @@ DEBUG_SET_ENV_VARS := \
 	export ELASTICSEARCH_VERIFY_CERTS=false; \
 	export ELASTICSEARCH_AWS_ACCESS_KEY_ID=debug; \
 	export ELASTICSEARCH_AWS_SECRET_ACCESS_KEY=debug; \
+	export ACTIVITY_STREAM_ACCESS_KEY_ID=some-id; \
+	export ACTIVITY_STREAM_SECRET_ACCESS_KEY=some-secret; \
+	export ACTIVITY_STREAM_IP_WHITELIST=1.2.3.4,2.3.4.5; \
```
richtier: this will cause the auth class to reject requests from local dev env, no? adding `127.0.0.1 should work.

activitystream/views.py
michalc: I'm a bit torn here: I'm unsure if adding this dependency really frees us from the responsibility in this respect.

- Ultimately, we still need to review the code and make sure it does what it needs to do. I think the most relavant is at https://github.com/un33k/django-ipware/blob/master/ipware/ip2.py#L24 (rather strangely/concerningly the repo also seems to include code from a previous version at https://github.com/un33k/django-ipware/blob/master/ipware/ip.py ?)

- So, I suspect adding the code from django-ipaware _increases_ the surface area for an attack. In PaaS, we will just be using the leftmost IP in X-Forwarded-For, and PaaS ensures that it can't be spoofed (from what I can tell by my previous contact with them). Just off the top of my head, were we to use django-ipaware....

    - The default behaviour is to use request.meta['HTTP_X_FORWARDED_FOR'] and then request.meta['X_FORWARDED_FOR'] (https://github.com/un33k/django-ipware/blob/master/ipware/defaults.py#L10). I'm not sure why there is  `X_FORWARDED_FOR` here because I thought that the `X-Forwarded-For` header maps to `HTTP_X_FORWARDED_FOR` https://docs.djangoproject.com/en/2.0/ref/request-response/#django.http.HttpRequest.META . It's _probably_ ok, but it's something to think about to make sure the loop at https://github.com/un33k/django-ipware/blob/master/ipware/ip2.py#L24 is all ok looping over both of them, they're in the correct order, etc.

    - If somehow an attacker managed to get X-Forwarded-For to not be passed or be the empty string, then they would be able to use one of the other headers at https://github.com/un33k/django-ipware/blob/master/ipware/defaults.py#L9 that could be passed unchanged. We would override the defaults, but then there would be far less point in using the library?

    - Similarly, there seems to be a rather complex way of accessing `request.META` https://github.com/un33k/django-ipware/blob/master/ipware/utils.py#L66 , using two dictionary `get`s with defaults, and replacing `_` with `-`. It's _probably_ ok, at worst it seems a bit needless, but all of a sudden there are a number of cases to think about. Say, we should make sure that the header `X_Forwarded-For` isn't used in preference to `X-Forwarded-For` for example. I think some(?) intermediate servers usually do not allow underscores in http headers names these days, but it's another layer to think about in something that's already quite complex.

- Being even more paranoid, we would be then dependant on another author (/someone who has managed to get their credentials). Pinning version of course mitigates this (and I think PyPy doesn't allow replacement of a version), but I think if bumping a version of a dependency, we typically we wouldn't re-review all of it's code(?).

There is the issue of local dev. However, using Postman / similar, you can just pass the X-Forwarded-For header to simulate what PaaS does. It's a bit more awkward of course, but a _slightly_ more awkard development environment I think is worth not having the above concerns, and having simpler security-critical code in production?

Happy to chat offline/in person of course.

activitystream/views.py
michalc: Done!

activitystream/views.py
michalc: As discussed offline, using a different approach since using Authorization header.

activitystream/views.py
michalc: Done!

makefile
michalc: Discussion relating to this in the comment above about django-ipaware

PR #419, opened 2018-06-15
[DNM] Support "verified with gov verify"

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-118)


supplier/models.py
```diff
@@ -52,6 +52,12 @@ class Supplier(TimeStampedModel):
         ),
     )
     is_company_owner = models.BooleanField(default=False)
+    govuk_verify_address = models.CharField(
```
richtier: this is the address of the user when they were verified with gov uk verify (personal data).

We can justify this with GDPR because it's kept for security purposes: we use the address and their name to identify the individual if they are a bad actor.

richtier:
we no longer need this functionality

PR #422, opened 2018-07-10
X-Forwarded-For spoofing mitigation

michalc:
We don't have network-level firewall capabilities, but in PaaS, the
second-to-last element in the X-Forwarded-For header cannot be spoofed.

### DIFF
--- a/activitystream/views.py
+++ b/activitystream/views.py
@@ -18,7 +18,7 @@ NO_CREDENTIALS_MESSAGE = 'Authentication credentials were not provided.'
 INCORRECT_CREDENTIALS_MESSAGE = 'Incorrect authentication credentials.'
 
 
-def _lookup_credentials(access_key_id):
+def lookup_credentials(access_key_id):
     """Raises a HawkFail if the passed ID is not equal to
     settings.ACTIVITY_STREAM_ACCESS_KEY_ID
     """
@@ -35,7 +35,7 @@ def _lookup_credentials(access_key_id):
     }
 
 
-def _seen_nonce(access_key_id, nonce, _):
+def seen_nonce(access_key_id, nonce, _):
     """Returns if the passed access_key_id/nonce combination has been
     used within 60 seconds
     """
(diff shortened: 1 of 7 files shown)


activitystream/views.py
```diff
@@ -30,83 +35,124 @@ def lookup_credentials(access_key_id):
...
-        return (ActivityStreamUser(), None)
+            logger.warning('Failed authentication {e}'.format(
+                e=e,
+            ))
+            raise AuthenticationFailed(INCORRECT_CREDENTIALS_MESSAGE)
+
+        return (None, hawk_receiver)
+
+
+class _ActivityStreamHawkResponseMiddleware:
```
richtier: why so many underscores? it's not a naming convention we use elsewhere.

activitystream/views.py
```diff
@@ -30,83 +35,124 @@ def lookup_credentials(access_key_id):
...
+            logger.warning(
+                'Failed authentication: the X-Forwarded-For header does not '
+                'contain enough IP addresses'
+            )
+            raise AuthenticationFailed(INCORRECT_CREDENTIALS_MESSAGE)
 
-        remote_address = \
-            request.META['HTTP_X_FORWARDED_FOR'].split(',')[0].strip()
+        # PaaS appends 2 IPs, where the IP connected from is the first
+        remote_address = ip_addesses[-2].strip()
```
richtier: this is strongly coupling this feature to the PaaS environment.

that could be solved by creating a swappable `RemoteIPAddressRetriver` and set it in settings:

```
# settinns.py
REMOTE_IP_ADDRESS_RETRIEVER = 'core.helpers.GovukPaaSRemoteIPAddressRetriver'
```

michalc: This is true, however, I believe we're slowly moving away from Heroku to PaaS, no plan to move away from PaaS, and locally you can add the headers when making requests.

It was already depending on this header, and I'm a bit cautious against making it configurable when there isn't a production use case for it, and it would be fairly easy to silently misconfigure.

Happy to add if you do feel strongly about it however.

activitystream/views.py
michalc: I thought it was a typical python convention: only `ActivityStreamViewSet` should be used outside of the module. However, after chatting with @konradko, will remove them.

activitystream/views.py
richtier: > no plan to move away from PaaS
It looks that way, but we cannot guarantee it. The easier we can move between environments with the fewest gotchas the happier future us/future devs will be when moving "frictionless" between dev/stage/prod or migrating to a different platform.

> would be fairly easy to silently misconfigure.
true. we can safeguard against that by using production setting as the default and making sure PRs are reviewed to avoid accidental changes.

> Happy to add if you do feel strongly about it however.
Smashing, thanks :)

activitystream/views.py
richtier: you could add `__all__ = ['ActivityStreamViewSet']` to the module, but I'm not sure if that would complicate testing.

michalc: Have removed the underscores

activitystream/views.py
michalc: Done!

(Also: not quite sure where to put things... happy to move them about)

PR #431, opened 2018-07-25
(tests) A larger range of higher level tests

michalc:
This is to try to track down the staging issue of not being able to dermine the
IP of the client, because it seems to hit the error of

    Not enough IP addresses in X-Forwarded-For

So far, no joy

### DIFF
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -19,11 +19,8 @@ class AdminIPRestrictorMiddleware(
     MESSAGE_UNABLE_TO_DETERMINE_IP_ADDRESS = 'Unable to determine remote IP'
 
     def get_ip(self, request):
-        return helpers.RemoteIPAddressRetriver().get_ip_address(request)
-
-    def process_request(self, request):
         try:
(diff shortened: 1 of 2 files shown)


core/middleware.py
```diff
@@ -14,11 +14,8 @@ class AdminIPRestrictorMiddleware(middleware.AdminIPRestrictorMiddleware):
     MESSAGE_UNABLE_TO_DETERMINE_IP_ADDRESS = 'Unable to determine remote IP'
 
     def get_ip(self, request):
-        return helpers.RemoteIPAddressRetriver().get_ip_address(request)
-
-    def process_request(self, request):
         try:
-            return super().process_request(request)
+            return helpers.RemoteIPAddressRetriver().get_ip_address(request)
```
richtier: I think raising the 404 is better placed in `proecess_request` because as a reader of the code I would expect a function named `get_ip` to just do stuff related to getting the IP. I would not expect raising 404 to be part of getting an IP address.

PR #459, opened 2019-02-19
Make sole trader enrolment work end to end

richtier:


enrolment/serializers.py
```diff
@@ -30,19 +33,22 @@ class Meta:
...
-        queryset = models.PreVerifiedEnrolment.objects.filter(
-            company_number=validated_data['number'],
-            email_address=validated_data['email_address'],
-            is_active=True
-        )
-        validated_data['verified_with_preverified_enrolment'] = (
-            queryset.exists()
-        )
+        if validated_data['company_type'] == Company.COMPANIES_HOUSE:
+            queryset = models.PreVerifiedEnrolment.objects.filter(
```
depsiatwal: do we need to update the tests to check rollback is handled appropriately  ?

depsiatwal: Also any update required to the enrolment/test_serializers.py as a result of these changes ?

PR #460, opened 2019-02-22
Create companies from spreadsheet

richtier:
[this task](https://uktrade.atlassian.net/browse/TT-1022)

ukan-import

### DIFF
--- a/company/forms.py
+++ b/company/forms.py
@@ -1,0 +1,194 @@
+import csv
+import io
+import json
+
+from directory_components.fields import PaddedCharField
+
+from django import forms
+from django.db import transaction
+
+from company import models
+
+
+class CompanyNumberField(PaddedCharField):
+
+    def __init__(self, fillchar='0', *args, **kwargs):
+        super().__init__(fillchar=fillchar, *args, **kwargs)
+
+    def to_python(self, value):
+        value = super().to_python(value)
+        if value:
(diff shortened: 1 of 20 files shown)


company/forms.py
```diff
@@ -0,0 +1,194 @@
...
+        super().__init__(fillchar=fillchar, *args, **kwargs)
+
+    def to_python(self, value):
+        value = super().to_python(value)
+        if value:
+            if 'n/a' in value.lower() or 'there is no' in value.lower():
+                return None
+            value = value.replace(' ', '')
+        else:
+            value = None
```
sdonk: Is value already going to be None in this branch?

company/forms.py
```diff
@@ -0,0 +1,194 @@
...
+        return self.address_lines[0] if self.is_address_parsable else ''
+
+    @property
+    def line_2(self):
+        return self.address_lines[1] if self.is_address_parsable else ''
+
+    @property
+    def po_box(self):
+        if self.is_address_parsable:
+            for line in self.address_lines:
```
sdonk: use `filter` to reduce cognitive complexity?

company/forms.py
richtier: it might also be an empty string.

the CSV file reader has no sense of `None`, so it's coerced to `None` here`

PR #476, opened 2019-03-26
Handle sole traders when retrieveing missing company data

richtier:


company/helpers.py
```diff
@@ -33,31 +31,12 @@ def generate_verification_code():
...
-    Raises:
-        KeyError: The companies house api may change the name of it's fields.
-        requests.exceptions.HTTPError: companies house may return non-200.
-        requests.exceptions.RequestException: A network error could occur.
-        ValueError: The date format companies house returned may be unexpected.
-        ValueError: The companies house api may return invalid json.
-
-    """
-
+def get_companies_house_profile(number):
```
depsiatwal: why remove the comments , they not applicable :-)

richtier: the nature of the function changed from `retrieve the profile...and parse the date` to simply `retrieve the profile`, therefore the docstring was not useful anymore.

I did not write a new docstring because the function is rather simple :)

PR #499, opened 2019-05-01
T-1376 allow searching expertise fields

depsiatwal:


company/tests/test_helpers.py
```diff
@@ -210,3 +212,56 @@ def test_address_parser(raw_address, line_1, line_2, po_box, postal_code):
...
+        choices.INDUSTRIES,
+        expertise_industries
+    )
+
+    choices_helper.extract_expertise_values(
+        expertise_values,
+        choices.EXPERTISE_REGION_CHOICES,
+        expertise_regions
+    )
+    choices_helper.extract_expertise_values(
```
richtier: why are these being called and nothing being done with the return value?

depsiatwal: ahh more of a private function in helpers 
removed from tests wrapper function will test this anyway

should make more sense now :-)

company/helpers.py
```diff
@@ -22,8 +22,53 @@
...
+    def get_expertise_values_list(
+        cls,
+        expertise_languages,
+        expertise_industries,
+        expertise_regions,
+        expertise_countries,
+    ):
+        expertise_values = []
+        cls._extract_expertise_values(
+            expertise_values,
```
richtier: in general when there are more than one argument in the function please use keyword arguments i.e., 
```
        cls._extract_expertise_values(
            expertise_values=expertise_values,
            lookup_choices=choices.INDUSTRIES,
            lookup_values=expertise_industries
        )
```

it's not great that `_extract_expertise_values` mutates the `expertise_values` list. it's mutating something it does not control. 

also, we're extracting attributes of the `company` and then bringing them back together again. we could simplify that a bit with a small modification:


```
class CompanyExpertiseFormatter:

   def __init__(self,  company):
       self.company = company

   def values_to_labels(self, available_options, chosen_values):
       # convert the list of chosen_values to a list of labels
        available_options_map = dict(available_options)
        labels = []
        for value in chosen_values:
            if value in available_options_map:
                labels = available_options_map[value]
         return labels

    @property
    def industry_labels(self):
        return self.values_to_labels(
            available_options=choices.INDUSTRIES,
            chosen_values=company.expertise_countries,
        )

    @property
    def language_labels(self):
        return self.values_to_labels(
            available_options=choices.EXPERTISE_LANGUAGES,
            chosen_values=company.expertise_languages,
        )
    
    @property
    def labels(self):
        return self.language_labels + self.industry_labels

```

then we can do

```
expertise_formatter = CompanyExpertiseFormatter(company)
...
expertise_labels=label_formatter.labels
```

it's easier to read, and avoids mutation of external variables

PR #506, opened 2019-05-09
TT-1317 import changes

depsiatwal:


company/forms.py
```diff
@@ -252,34 +255,44 @@ def clean_csv_file(self):
...
                 )
-            else:
-                companies = models.Company.objects.filter(name=data['name'])
 
             if companies.count() == 0:
                 self.add_bulk_errors(
                     errors=self.update_errors,
-                    row_number=i,
-                    line_errors=self.MSG_COMPANY_NOT_FOUND),
+                    row_number=i+2,
```
richtier: why +2?

depsiatwal: first one is header and it probably starts at 0. 
Hence +2 reflects the actual excel row

PR #517, opened 2019-05-14
Fixed pre-verified enrolments search throwing server error

gauravdott:


fixtures/development.json
```diff
@@ -159,20 +193,5 @@
         "company": 32,
         "slug": "example-company"
     }
-},
-{
```
richtier: this user existing adds value - SSO also has `development.json` fixture that has a user with that email and sso id, meaning the example development user has a company "out of the box"

PR #518, opened 2019-05-14
Fix for Fixed pre-verified enrolments search throwing server error

gauravdott:


enrolment/tests/test_models.py
```diff
@@ -7,9 +7,11 @@
 
 
 def test_trusted_source_signup_code_str():
-    instance = models.PreVerifiedEnrolment(email_address='jim@example.com')
+    instance = models.PreVerifiedEnrolment(company_number='12345678',
+                                           email_address='jim@example.com')
 
-    assert str(instance) == 'jim@example.com'
+    assert str(instance) == '12345678'
+    assert type(instance) is not None
```
richtier: `assert instance is not None` :)

also, the line above checking it's equal to `'12345678'` makes this check that it's not None redundant so I would remove the `not None` check :)

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 - [TT-1408](https://uktrade.atlassian.net/browse/TT-1408) Customize ISD search results order.
 
 ### Fixed bugs:
+- [TT-7](https://uktrade.atlassian.net/browse/TT-7) Fixed Server Error (500) when searching for pre-verified enrolments
```
richtier: please remove the links to jira

could you please also clear up the changelog so all jira links are removed :pray:

enrolment/tests/test_models.py
gauravdott: ok, makes sense.

CHANGELOG.md
gauravdott: ok

PR #523, opened 2019-05-16
staging release

richtier:


richtier:
let's release this after the 1pm launch

PR #536, opened 2019-06-17
TT-1558 Mask prod data

depsiatwal:


core/tests/test_management_commands.py
```diff
@@ -3,6 +3,31 @@
...
+
+
+@pytest.fixture()
+def new_companies():
+    new_companies = []
+    for _ in range(10):
+        new_companies.append(
+            CompanyFactory()
+        )
+    return new_companies
```
richtier: `return CompanyFactory.build_batch(10)` :)

PR #537, opened 2019-06-19
Tt 1491 sort alphabetically with equal scores

gauravdott:


company/search.py
```diff
@@ -71,7 +71,10 @@ class CompanyDocument(Document):
     logo = field.Keyword(index=False, store=True)
     has_single_sector = field.Boolean()
     modified = field.Date(index=False)
-    name = field.Text(copy_to='wildcard')
+
+    ordering_name = field.Keyword()
+    name = field.Text(copy_to=['wildcard', 'ordering_name'], boost=2)
+
```
richtier: please remove extra new lines.

company/views.py
```diff
@@ -323,6 +323,18 @@ def get(self, request, *args, **kwargs):
             .search()
             .filter('term', is_published_investment_support_directory=True)
             .query(query)
+            .sort(
+                {
+                    "_score": {
+                        "order": "desc"
+                    }
```
richtier: this is more readable:
```
        .sort({
            {"_score": {"order": "desc"}},
            {"ordering_name": {"order": "asc"}
         })
```

company/search.py
gauravdott: Done!

company/views.py
gauravdott: Done.

company/helpers.py
```diff
@@ -187,9 +187,27 @@ def build_search_company_query(params):
...
+                'bool',
+                must=must,
+                should=should,
+                minimum_should_match=1 if should else 0
+            ),
+            functions=[
+                SF({
+                    'weight': 5,
+                    'filter': Q('match', name=term)
+                })
```
richtier: can you please also add `match_phrase` too in order to support partial matches across words. something like
```
                SF({
                    'weight': 5,
                    'filter': Q(
                        'bool',
                        should=[
                            Q('match_phrase', name=term)),
                            Q('match', name=term)),
                       ]
                   }),
```

gauravdott: Done!

PR #545, opened 2019-07-02
Mask prod date v2

depsiatwal:
 - [x] Changelog entry added.


supplier/tests/factories.py
```diff
@@ -10,8 +10,8 @@ class SupplierFactory(factory.django.DjangoModelFactory):
     name = factory.Faker('name', locale='en_GB')
     mobile_number = factory.fuzzy.FuzzyText(length=11, chars='1234567890')
     company_email = factory.LazyAttribute(
-        lambda supplier: '%s@example.com' % supplier.name.replace(
-            ',', '_').replace(' ', '_')
+        lambda supplier: '%s@example.com' % (supplier.name.replace(
+            ',', '_').replace(' ', '_') + str(supplier.sso_id))
```
richtier: you can replace `.replace(',', '_').replace(' ', '_') ` with [slugify](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.text.slugify)

```
factory.LazyAttribute(lambda x: f'({slugify(x.name)}-{x.sso_id}@example.com')

PR #558, opened 2019-07-19
send letter via govnotify

depsiatwal:

 - [x] Changelog entry added.
 - [x] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.


company/tests/test_signals.py
```diff
@@ -103,7 +103,10 @@ def test_letter_sent(mock_send_letter, settings):
     settings.FEATURE_VERIFICATION_LETTERS_ENABLED = True
     company = factories.CompanyFactory(verification_code='test')
 
-    mock_send_letter.assert_called_with(company=company)
+    mock_send_letter.assert_called_with(
```
richtier: please avoid using `assert_called_with` and if you hit code that uses it, replace with
```
assert mock_send_letter.call_count == 1
assert mock_send_letter.call_args == mock.call(...)
```

[reason](https://engineeringblog.yelp.com/2015/02/assert_called_once-threat-or-menace.html)

conf/.env2
```diff
@@ -0,0 +1,7 @@
+GOVNOTIFY_VERIFICATION_LETTER_TEMPLATE_ID=22d1803a-8af5-4b06-bc6c-ffc6573c4c7d
+DIRECTORY_FORMS_API_API_KEY=d7VlRSQ7ervPu3PBzYg3A80l016dFu3SJ7AMgkrdGpUIdUyZeA6ACBsCl3yWAw75
+DIRECTORY_FORMS_API_SENDER_ID=f51d396d-a898-4415-b165-03d70a5544bc
+
+export DIRECTORY_FORMS_API_API_KEY=d7VlRSQ7ervPu3PBzYg3A80l016dFu3SJ7AMgkrdGpUIdUyZeA6ACBsCl3yWAw75; \
+	export DIRECTORY_FORMS_API_SENDER_ID=f51d396d-a898-4415-b165-03d70a5544bc; \
+	export GOVNOTIFY_VERIFICATION_LETTER_TEMPLATE_ID=22d1803a-8af5-4b06-bc6c-ffc6573c4c7d; \
```
richtier: please remove this file

makefile
```diff
@@ -85,6 +85,10 @@ DEBUG_SET_ENV_VARS := \
 	export IP_RESTRICTOR_REMOTE_IP_ADDRESS_RETRIEVER=ipware; \
 	export SOLE_TRADER_NUMBER_SEED=32; \
 	export DIRECTORY_CONSTANTS_URL_GREAT_DOMESTIC=http://profile.trade.great:8006/; \
+	export DIRECTORY_FORMS_API_BASE_URL=http://forms.trade.great:8011; \
+	export DIRECTORY_FORMS_API_API_KEY=debug_key; \
+	export DIRECTORY_FORMS_API_SENDER_ID=debug_key; \
```
richtier: please move `DIRECTORY_FORMS_API_API_KEY` and `DIRECTORY_FORMS_API_SENDER_ID` from `DEBUG_SET_ENV_VARS` to `TEST_SET_ENV_VARS` (so the tests dont fail)

we want to remove them from `DEBUG_SET_ENV_VARS` so devs are not tempted to set real values in the makeifle then accidentally commit the secrets. It's happened in the past and .env file is a strong way to avoid that.

As such, please and add a README.md entry indicating that the dev needs to set in their `conf/.env` file. Can steal from [here](https://github.com/uktrade/directory-forms-api#configuration)

also I think directory-api does not yet utilise `.env` file so will need to add [this](https://github.com/uktrade/directory-forms-api/blob/develop/conf/settings.py#L11) to settings

PR #563, opened 2019-07-24
TT-1619 added registration letter

depsiatwal:

 - [x] Changelog entry added.


company/apps.py
```diff
@@ -19,6 +19,10 @@ def ready(self):
             receiver=signals.send_first_verification_letter,
             sender='company.Company'
         )
+        post_save.connect(
+            receiver=signals.send_first_registration_letter,
```
richtier: is there a second registration letter? if not, could call this `send_registration_letter`

company/migrations/0086_auto_20190722_1759.py
```diff
@@ -0,0 +1,25 @@
...
+
+    operations = [
+        migrations.AddField(
+            model_name='company',
+            name='date_registration_letter_sent',
+            field=models.DateTimeField(blank=True, null=True),
+        ),
+        migrations.AddField(
+            model_name='company',
+            name='is_registration_letter_sent',
```
richtier: do we need this? `date_registration_letter_sent` will be empty if the letter was not sent

richtier: oh I see we do similar thing for `is_verification_letter_sent` :+1:

company/tests/test_signals.py
```diff
@@ -43,6 +53,59 @@ def test_sends_verification_letter_post_save(settings):
...
+    'is_registration_letter_sent, '
+    'has_valid_address, company_type',
+    [
+        [False, True,  True,  company_types.COMPANIES_HOUSE],
+        [True,  True,  False, company_types.COMPANIES_HOUSE],
+        [True,  True,  True,  company_types.CHARITY],
+        [True,  True,  True,  company_types.PARTNERSHIP],
+        [True,  True,  True,  company_types.SOLE_TRADER],
+    ]
+)
```
richtier: I recently discovered we can actually do nested paramterize:
```
@pytest.mark.parametrize('letter_registration_enabled', (True, False))
@pytest.mark.parametrize('has_valid_address', (True, False))
@pytest.mark.parametrize('company_type', (
    company_types.COMPANIES_HOUSE,
    company_types.CHARITY,
    company_types.PARTNERSHIP,
    company_types.SOLE_TRADER,
))
```

PR #564, opened 2019-07-29
TT-851 added company house status

depsiatwal:

 - [x] Changelog entry added.


company/management/commands/retrieve_companies_house_company_status.py
```diff
@@ -0,0 +1,33 @@
...
+            company_type=company_types.COMPANIES_HOUSE
+        )
+        failed = 0
+        succeded = 0
+        for company in queryset:
+            try:
+                profile = helpers.get_companies_house_profile(company.number)
+                if profile.get('company_status'):
+                    if profile.get('company_status').lower() == 'active':
+                        company.companies_house_company_status = 'active'
```
richtier: there are some company statuses other than `action` that are...active

there is one that means "they're being timeboxes 6 months to get their mess together and try to become profitable".

It's worth going through the list of statuses and see which ones are "pseudo active"

PR #565, opened 2019-07-31
(fix) redirects in API with tests

agiamas:
 - [X] Changelog entry added.

Back end changes for automating redirects. FE changes will follow in a PR in great domestic UI.


conf/urls.py
```diff
@@ -245,6 +246,11 @@
         enrolment.views.PreverifiedCompanyView.as_view(),
         name='enrolment-preverified'
     ),
+    url(
+        r'^redirects/urls/(?P<key>.*)/$',
```
richtier: we could simplify this to `redirects/(?P<path>.*)/`

redirects/models.py
```diff
@@ -0,0 +1,14 @@
+from django.db import models
+from redirects.validators import self_redirect
+
+
+class Redirect(models.Model):
+    class Meta:
+        ordering = ['id']
+
+    id = models.AutoField(auto_created=True, primary_key=True)
```
richtier: django adds `id` for us, no need to explicitly set

redirects/models.py
```diff
@@ -0,0 +1,14 @@
...
+    source_url = models.CharField(max_length=200, blank=True, null=True)
```
richtier: would it be worth making this `unique=True`?

redirects/serializers.py
```diff
@@ -0,0 +1,13 @@
...
+from redirects import models
+
+
+class RedirectSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.Redirect
+        fields = (
+            'id',
+            'source_url',
+            'target_url',
```
richtier: I think we only need to expose the `target_url`. The requestor already has the `source_url` and the primary key is not needed too as far as I can tell

redirects/validators.py
```diff
@@ -0,0 +1,12 @@
+from django.core.exceptions import ValidationError
+import environ
+
+env = environ.Env()
+
+NO_SELF_REDIRECT_ALLOWED_MESSAGE = 'Can not redirect to self.'
+DOMAIN = env.str('DIRECTORY_CONSTANTS_URL_EXPORT_READINESS')
```
richtier: we already have access to this value: `directory_constants.urls.SERVICES_GREAT_DOMESTIC`

however, `SERVICES_GREAT_DOMESTIC` starts with http/https (depending on environment) and `target_url` may be http or https

we can do something like
```
from urllib.parse import urlparse

great_domestic_ui = urlparse(directory_constants.urls.SERVICES_GREAT_DOMESTIC)
target = urlparse(target_url)
if target.netloc == great_domestic_ui.netloc:
    raise ...
```

redirects/validators.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+def self_redirect(target_url):
```
richtier: something like `is_not_great_domestic_url` would be more apt

redirects/views.py
```diff
@@ -0,0 +1,21 @@
+from rest_framework.views import APIView
+from rest_framework.response import Response
+
+from redirects import models, serializers
+from django.http import HttpResponseNotFound
+
+
+class RetrieveRedirect(APIView):
```
richtier: I think we can simplify to:

```
from rest_framework import generics

class RetrieveRedirectRetrieveAPIView(generics.RetrieveAPIView):
    serializer_class = serializers.RedirectSerializer
    queryset = models.Redirect.objects.objects.all()
    lookup_field = 'source_url'
    lookup_url_kwarg = 'key'
```

redirects/views.py
```diff
@@ -0,0 +1,21 @@
...
+
+    permission_classes = []
+
+    def get(self, request, *args, **kwargs):
+        serializer_class = serializers.RedirectSerializer
+        source_url = '/'+kwargs['key']+'/'
```
richtier: from the validator I got the impression the `source_url` would be an absolte url e.g, "http://www.google.com", but here we're prefixing with `/`, resulting in `/http://www.google.com`

if we do the urls.py change I suggested then we won't need to manually append '/' )

agiamas:
won't implement at this time. Will implement in a simpler single tier architecture.

PR #566, opened 2019-08-01
TT-1697 added SSO

depsiatwal:
 - [x] Changelog entry added.


conf/settings.py
```diff
@@ -166,6 +169,22 @@
...
+        'authbroker_client.backends.AuthbrokerBackend'
+    ]
+
+    LOGIN_URL = reverse_lazy('authbroker_client:login')
+    LOGIN_REDIRECT_URL = reverse_lazy('admin:index')
+
+# authbroker config
+AUTHBROKER_URL = env.str('STAFF_SSO_AUTHBROKER_URL')
+AUTHBROKER_CLIENT_ID = env.str('AUTHBROKER_CLIENT_ID')
+AUTHBROKER_CLIENT_SECRET = env.str('AUTHBROKER_CLIENT_SECRET')
```
richtier: should these only be set if `FEATURE_ENFORCE_STAFF_SSO_ENABLED` is enabled? if so consider putting them under the `if FEATURE_ENFORCE_STAFF_SSO_ENABLED`

core/middleware.py
```diff
@@ -9,6 +11,16 @@ class SignatureCheckMiddleware(
...
+            'admin', 'healthcheck', 'authbroker_client'
+        ] or request.path_info.startswith('/admin/login'):
             return False
         return super().should_check(request)
+
+
+class AuthenticatedUserPermissionCheckMiddleware(MiddlewareMixin):
+
+    def process_view(self, request, view_func, view_args, view_kwarg):
+        if request.user is not None:
```
richtier: will this not stop directory-ui-supplier and directory-sso-profile from communicating with api via the endpoints?

usermanagement/management/commands/grant_staff_status.py
```diff
@@ -0,0 +1,26 @@
...
+from django.contrib.auth.models import User
+
+
+class Command(BaseCommand):
+    """Pass the username to grant staff status."""
+
+    help = 'Used to create a supplier and company for integration tests.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('username', nargs='+')
```
richtier: `parser.add_argument('username', nargs=1)` will result in `options['username']` being `'foo'` rather than `['foo']`

depsiatwal:
ahh my comments were in review state

core/middleware.py
richtier: soon we will be using the `SSOAuthenticationBackend` in API too - that will mean `request.sso_user` becomes `request.user`. More djangoey. that would break this.

I think Activity Stream may do something with `request.user` too.

Problem is solved by checking if the requested url is in admin namespace. If you rename this middleware to something like `AdminPermissionCheck`

core/middleware.py
```diff
@@ -9,6 +11,18 @@ class SignatureCheckMiddleware(
...
+        ] or request.path_info.startswith('/admin/login'):
             return False
         return super().should_check(request)
+
+
+class AdminPermissionCheckMiddleware(MiddlewareMixin):
+
+    def process_view(self, request, view_func, view_args, view_kwarg):
+        if request.user is not None and (
+                request.resolver_match.namespace == 'admin' or request.path_info.startswith('/admin/login')
```
richtier: bit hard to read. something like the following would be a bit easier

```
    is_admin_url = request.resolver_match.namespace == 'admin' or request.path_info.startswith('/admin/login')
    if is_admin_url and request.user is not None and request.user.is_authenticated:
        if not request.user.is_staff:
            return HttpResponse('User not authorized for admin access', status=401)
```

PR #577, opened 2019-08-27
Tt 1734 allow multiple users to be added as member

gauravdott:

 - [x] Changelog entry added.


supplier/views.py
```diff
@@ -108,3 +108,15 @@ def post(self, request, *args, **kwargs):
         supplier.role = user_roles.MEMBER
         supplier.save()
         return Response()
+
+
+class RegisterCollaboratorRequestView(CreateAPIView):
```
richtier: this is very related to the collaborator views in `company.view`. I would, therefore, move this view (and related serializers etc) to `company`. Simpler for devs to find if all collaborator stuff is in the same place.

richtier: `RegisterCollaboratorRequestView` is a bit of a misnoma.

this view is not registering a `CollaboratorRequest` (`company.models.CollaboratorRequest`) - it's creating a `Supplier`.

I think `AddCollaboratorView` would be more accurate.

please also rename the url name accordingly too :)

supplier/views.py
```diff
@@ -108,3 +108,15 @@ def post(self, request, *args, **kwargs):
...
+    serializer_class = serializers.RegisterCollaboratorRequestSerializer
+    permission_classes = []
+
+    def create(self, request, *args, **kwargs):
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        self.perform_create(serializer)
+        data = {'company_email': serializer.instance.company_email}
+        return Response(data, status=201)
```
richtier: what is this doing differently to the default behavior of `CreateAPIView.create`? I think we can remove this override and use the default behaviour

https://github.com/encode/django-rest-framework/blob/master/rest_framework/mixins.py#L16

supplier/serializers.py
```diff
@@ -57,3 +61,34 @@ class Meta:
...
+            company = Company.objects.get(number=data['company_number'])
+        except Company.DoesNotExist:
+            raise serializers.ValidationError({
+                '__all__': 'Company does not exist'
+            })
+        else:
+            data['company'] = company.pk
+
+        if not data.get('role', ''):
+            data['role'] = user_roles.MEMBER
```
richtier: we can do this without custom code: 
```
class Meta
    extra_kwargs = {
        'role': {'default': user_roles.MEMBER},
    }
```

supplier/serializers.py
```diff
@@ -57,3 +61,34 @@ class Meta:
...
+
+    def to_internal_value(self, data):
+        if isinstance(data, QueryDict):
+            data = data.dict()
+        try:
+            company = Company.objects.get(number=data['company_number'])
+        except Company.DoesNotExist:
+            raise serializers.ValidationError({
+                '__all__': 'Company does not exist'
+            })
```
richtier: we can avoid custom code here:
```

class RegisterCollaboratorRequestSerializer(...):
    company = serializers.SlugRelatedField(slug_field='company_number')
```

this will mean that sso-profile will need to provide the company number as `{company: <company_number>, ...}` instead of `{company_number: <company_number>, ...}`.

supplier/serializers.py
```diff
@@ -57,3 +61,34 @@ class Meta:
...
+        else:
+            data['company'] = company.pk
+
+        if not data.get('role', ''):
+            data['role'] = user_roles.MEMBER
+
+        return super().to_internal_value(data)
```
richtier: by using `extra_kwargs` and `SlugRelatedField` we can avoid this `to_internal_value` override completely :)

supplier/views.py
gauravdott: We are just exposing the email on 201 rather than all the serializer data.

supplier/views.py
gauravdott: Makes sense

company/serializers.py
```diff
@@ -8,6 +8,7 @@
 from django.http import QueryDict
 
 from company import helpers, models, validators
+from company.models import Company
```
richtier: `company.models` is already imported. please use `models.Company`

PR #578, opened 2019-08-28
Tt 1743 company ltd stopwords

depsiatwal:

 - [x] Changelog entry added.


company/search_filters.py
```diff
@@ -1,5 +1,16 @@
...
+companies_stopwords_filter = analysis.token_filter(
+    'companies_stopwords',
+    type='stop',
+    stopwords=['limited', 'ltd', 'plc', 'llp', 'lp', 'rc', 'partnership', 'ngo']
+)
+
+english_stopwords_filter = analysis.token_filter(
+    'english_stopwords',
+    type='stop',
+    stopwords=['and', '&', '+', ',', 'the']
```
richtier: pretty sure these are included in the default list of stopwords. when I search for https://www.great.gov.uk/international/trade/search/?q=things+and+stuff `and` is ignored.

from the internet it seems the following are included by default:
```
"a", "an", "and", "are", "as", "at", "be", "but", "by",
"for", "if", "in", "into", "is", "it",
"no", "not", "of", "on", "or", "such",
"that", "the", "their", "then", "there", "these",
"they", "this", "to", "was", "will", "with"
```

company/tests/test_views.py
```diff
@@ -1420,6 +1478,26 @@ def test_search_results(
         assert hit['_id'] in expected
 
 
+@pytest.mark.rebuild_elasticsearch
+@pytest.mark.django_db
+@pytest.mark.parametrize('url', search_urls)
+@pytest.mark.parametrize('stop_term', ['limited', 'plc', 'ltd'])
+def test_search_results_stopwords(url, stop_term, search_companies_stopwords, api_client):
+    data = {
+        'term': 'mycompany {0}'.format(stop_term),
```
richtier: we can now do `f'mycompany {stop_term}'`

PR #579, opened 2019-08-29
Consolidate collaboration models in backwards compatible way

richtier:
 - [x] Changelog entry added.


richtier:
eventually, we will delete the old views and models

PR #587, opened 2019-09-03
Tt 1538 s3 paa s bucket setup

depsiatwal:
 - [ ] Changelog entry added.


conf/settings.py
```diff
@@ -361,21 +361,29 @@
 S3_USE_SIGV4 = env.bool('S3_USE_SIGV4', True)
 AWS_S3_HOST = env.str('AWS_S3_HOST', 's3.eu-west-1.amazonaws.com')
 
+AWS_S3_DEFAULT_BINDING_BUCKET_NAME = env.str('AWS_S3_DEFAULT_BINDING_BUCKET_NAME')
+AWS_S3_DATA_SCIENCE_BINDING_BUCKET_NAME = env.str('AWS_S3_DATASCIENCE_BINDING_BUCKET_NAME')
+AWS_STORAGE_BUCKET_NAME_DATA_SCIENCE = ''
```
richtier: we can move these env vars under `if 'aws-s3-bucket' in VCAP_SERVICES`

I suspect `AWS_STORAGE_BUCKET_NAME_DATA_SCIENCE` is not needed

conf/settings.py
```diff
@@ -361,21 +361,29 @@
...
+            AWS_S3_ENCRYPTION = True
+            AWS_DEFAULT_ACL = None
+        elif aws_paas_bucket['name'] == AWS_S3_DATA_SCIENCE_BINDING_BUCKET_NAME:
+            credentials = aws_paas_bucket['credentials']
+            AWS_ACCESS_KEY_ID_DATA_SCIENCE = credentials['aws_access_key_id']
+            AWS_SECRET_ACCESS_KEY_DATA_SCIENCE = credentials['aws_secret_access_key']
+            AWS_STORAGE_BUCKET_NAME_DATA_SCIENCE = credentials['bucket_name']
+            AWS_S3_REGION_NAME_DATA_SCIENCE = credentials['aws_region']
+            AWS_S3_ENCRYPTION_DATA_SCIENCE = True
+            AWS_DEFAULT_ACL_DATA_SCIENCE = None
```
richtier: the looping approach will hide an error if we do a typo in `AWS_S3_DEFAULT_BINDING_BUCKET_NAME` or `AWS_S3_DATA_SCIENCE_BINDING_BUCKET_NAME`:

the blocks that sets `AWS_ACCESS_KEY_ID` will never be hit if `aws_paas_bucket['name'] == AWS_S3_DEFAULT_BINDING_BUCKET_NAME` is not true. That can happen if we doa  typo in vault

this will result in error later when code tries to use those settings

we can make the error happen sooner at app start up by doing something like this:

```
bucket_credentials_map = {
    item['name']: item
    for item in VCAP_SERVICES['aws-s3-bucket']
}
default_credentials = bucket_credentials_map[AWS_S3_DEFAULT_BINDING_BUCKET_NAME]
datascience_credentials = bucket_credentials_map[AWS_S3_DATA_SCIENCE_BINDING_BUCKET_NAME]

```

makefile
```diff
@@ -88,7 +87,10 @@ DEBUG_SET_ENV_VARS := \
 	export CELERY_TASK_ALWAYS_EAGER=true; \
 	export AUTHBROKER_CLIENT_ID=debug; \
 	export AUTHBROKER_CLIENT_SECRET=debug; \
-	export STAFF_SSO_AUTHBROKER_URL=https://www.test.com
+	export STAFF_SSO_AUTHBROKER_URL=https://www.test.com; \
+	export AWS_S3_DEFAULT_BINDING_BUCKET_NAME=debug; \
+	export AWS_S3_DATASCIENCE_BINDING_BUCKET_NAME=debug
```
richtier: if we move the `AWS_S3_DEFAULT_BINDING_BUCKET_NAME` inside `if 'aws-s3-bucket' in VCAP_SERVICES` block then we dont need to set these in makefile as in local env s3 is not used

supplier/views.py
```diff
@@ -89,10 +89,16 @@ def get_queryset(self):
 
 
 class SupplierCSVDownloadAPIView(CSVDumpAPIView):
```
richtier: simplified :)
```
if settings.STORAGE_CLASS_NAME == 'default':
    class SupplierCSVDownloadAPIView(CSVDumpAPIView):
        bucket = settings.AWS_STORAGE_BUCKET_NAME_DATA_SCIENCE
        key = settings.SUPPLIERS_CSV_FILE_NAME
        filename = settings.SUPPLIERS_CSV_FILE_NAME
```

buyer/views.py
```diff
@@ -11,7 +11,9 @@ class BuyerCreateAPIView(CreateAPIView):
     permission_classes = []
 
 
-class BuyerCSVDownloadAPIView(CSVDumpAPIView):
-    bucket = settings.CSV_DUMP_BUCKET_NAME
-    key = settings.BUYERS_CSV_FILE_NAME
-    filename = settings.BUYERS_CSV_FILE_NAME
+if settings.STORAGE_CLASS_NAME == 'default':
+    # this view only works if s3 is in use (s3 is default. in local dev local storage is used
```
richtier: missing closing paran :)

core/tests/test_views.py
```diff
@@ -16,6 +16,13 @@ def reload_urlconf(urlconf=None):
         import_module(urlconf)
 
 
+def reload_module(module):
+    if module in sys.modules:
+        reload(sys.modules[module])
+    else:
+        import_module(module)
```
richtier: can you also update `reload_urlconf` to cal this function instead of duplicating the code :)

PR #588, opened 2019-09-09
fix datascience encryption

depsiatwal:
- [ ] Changelog entry added.


conf/settings.py
```diff
@@ -388,8 +388,8 @@
     AWS_SECRET_ACCESS_KEY_DATA_SCIENCE = credentials['aws_secret_access_key']
     AWS_STORAGE_BUCKET_NAME_DATA_SCIENCE = credentials['bucket_name']
     AWS_S3_REGION_NAME_DATA_SCIENCE = credentials['aws_region']
-    AWS_S3_ENCRYPTION_DATA_SCIENCE = True
-    AWS_DEFAULT_ACL_DATA_SCIENCE = None
+    AWS_S3_ENCRYPTION_DATA_SCIENCE = 'AES256'
+    AWS_DEFAULT_ACL_DATA_SCIENCE = 'public-read'
```
richtier: are these really settings? could we simply do
```
    s3.put_object(
        Bucket=bucket,
        Key=key,
        Body=file_object.getvalue()
        Body=file_object.getvalue(),
        ACL='public-read',
        ServerSideEncryption='AES256',
```

I think they're not settings because:
 - we're not likely to change them ever
 - we don't need different values depending on environments
 - we have fill control over the upload code, so we can hard code the values

PR #597, opened 2019-09-19
Allow company names longer than 250 chars

richtier:
 - [x] Changelog entry added.


notifications/tasks.py
```diff
@@ -15,12 +15,6 @@ def lock_acquired(lock_name):
     return cache.add(lock_name, 'acquired', 72000)
 
 
-@app.task
-def no_case_studies():
-    if lock_acquired('no_case_studies'):
-        notifications.no_case_studies()
```
richtier: this notification no longer exists. failing in sentry

PR #603, opened 2019-10-02
TT-1924 - allow to programmatically change verification flags via testapi

kowalcj0:
 - [x] Changelog entry added.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,8 +55,12 @@ jobs:
           name: Run tests
           command: |
             . venv/bin/activate
-            make pytest -- --cov=. --cov-config=.coveragerc --cov-report=term
+            make pytest -- --cov=. --cov-config=.coveragerc --cov-report=term --junitxml=test-reports/junit.xml
             codecov
+      - store_test_results:
+          path: test-reports
+      - store_artifacts:
+          path: test-reports
   flake8:
     docker:
       - image: circleci/python:3.6.6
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -55,8 +55,12 @@ jobs:
           name: Run tests
           command: |
             . venv/bin/activate
-            make pytest -- --cov=. --cov-config=.coveragerc --cov-report=term
+            make pytest -- --cov=. --cov-config=.coveragerc --cov-report=term --junitxml=test-reports/junit.xml
```
richtier: I think we will need to run `            mkdir test-reports` before `make pytest` for this change to work

kowalcj0: :+1:

testapi/views.py
```diff
@@ -55,14 +60,37 @@ def get(self, request, *args, **kwargs):
...
+            company.verified_with_identity_check = data.get('verified_with_identity_check')
+            should_save = True
+        if 'verified_with_code' in data:
+            company.verified_with_code = data.get('verified_with_code')
+            should_save = True
+        if should_save:
+            company.save()
+            return Response(status=204)
+        else:
+            return Response(status=200, data={"message": "Nothing to save"})
```
richtier: why not do this with a serializer?
```
serializer = TheSerializer(instance=company, data=request.data, partial=True)
serializer.is_valid(raise=True)
serializer.save()
```

`TheSerializer` would allow `verified_with_identity_check` or `verified_with_code` to be set.

kowalcj0: I didn't know such thing was possible :)
on it.

kowalcj0: done

PR #608, opened 2019-10-16
TT-2012 add non-ch verification emails

depsiatwal:

 - [x] Changelog entry added.


depsiatwal:
image

PR #624, opened 2019-11-18
Tt 1608 member request admin

depsiatwal:
 - [x] Changelog entry added.


company/views.py
```diff
@@ -238,7 +238,7 @@ def get_permissions(self):
         elif self.action == 'partial_update':
             permission_classes = [IsAuthenticatedSSO]
         else:
-            permission_classes = [IsAuthenticatedSSO, permissions.IsCompanyAdmin]
+            permission_classes = [IsAuthenticatedSSO]
```
richtier: why this change? this will allow any logged in user (even humble members) to create and delete collaboration invites. not good :)

the logic in `develop` is:
- allow anyone to retrieve a collaboration invite
- allow any logged in user to update a collaboration invite
- allow ony the company admin to create or delete the collaboration invite

PR #633, opened 2019-12-02
Fix enrolment

richtier:

 - [x] Changelog entry added.


company/signals.py
```diff
@@ -23,12 +23,13 @@ def send_first_verification_letter(sender, instance, *args, **kwargs):
...
 
 
-def send_company_claimed_letter(sender, instance, *args, **kwargs):
+def send_company_registration_letter(sender, instance, *args, **kwargs):
     should_send_letter = all([
         settings.FEATURE_REGISTRATION_LETTERS_ENABLED,
         not instance.is_registration_letter_sent,
         instance.company_type == company_types.COMPANIES_HOUSE,
         bool(instance.address_line_1 and instance.postal_code),
+        instance.company_users.exists(),
```
depsiatwal: does that mean if in the race condition now if user doesn't exist 
then a letter isn't sent ?

richtier: this prevents the letter from being sent until the company has a user.

note we call this signal explicitly in the enrolment view after the company user is created: https://github.com/uktrade/directory-api/pull/633/files#diff-c582450649e17f37ef3b939b1fc79d13R36

PR #664, opened 2020-01-16
Cleaned up obseolete settings

richtier:
 - [x] Changelog entry added.


notifications/email.py
```diff
@@ -90,20 +90,6 @@ def send(self):
         )
 
 
-class NoCaseStudiesNotification(SupplierNotificationBase):
```
richtier: this is dead code

PR #679, opened 2020-02-10
Mvp 108 export plan saving

depsiatwal:
 - [x] Changelog entry added.


exportplan/views.py
```diff
@@ -0,0 +1,32 @@
...
+from exportplan import serializers, models
+from exportplan.models import CompanyExportPlan
+
+
+class CompanyExportPlanRetrieveUpdateView(generics.RetrieveUpdateAPIView):
+    serializer_class = serializers.CompanyExportPlanSerializer
+    queryset = CompanyExportPlan.objects.all()
+    lookup_field = 'pk'
+
+    def get_permissions(self):
```
richtier: can replace this with `permission_classes = [IsAuthenticatedSSO]`

conf/urls.py
```diff
@@ -208,6 +209,16 @@
         exporting.views.RetrieveOfficesByPostCode.as_view(),
         name='offices-by-postcode'
     ),
+    url(
+        r'^exportplan/company-export-plan/$',
+        exportplan.views.CompanyExportPlanListCreateAPIView.as_view(),
+        name='export-plan-list-create'
+    ),
+    url(
+        r'^exportplan/company-export-plan/(?P<pk>.*)/$',
```
richtier: I think `(?P<pk>.*)` should be replaced with `(?P<pk>[0-9]+)`

the current one is very open: any character or none. I bet we will get unexpected side effects if we have the pattern matching too open

exportplan/views.py
```diff
@@ -0,0 +1,29 @@
...
+    lookup_field = 'pk'
+
+
+class CompanyExportPlanListCreateAPIView(generics.ListCreateAPIView):
+    serializer_class = serializers.CompanyExportPlanSerializer
+    queryset = CompanyExportPlan.objects.all()
+    lookup_field = 'sso_id'
+
+    def get_permissions(self):
+        permission_classes = [IsAuthenticatedSSO]
```
richtier: `permission_classes = [IsAuthenticatedSSO]` here too :)

PR #683, opened 2020-02-14
add export plan data

depsiatwal:
 - [ ] Changelog entry added.


exportplan/models.py
```diff
@@ -10,6 +10,7 @@ class CompanyExportPlan(TimeStampedModel):
     company = models.ForeignKey(
         Company, related_name='company_export_plans', on_delete=models.CASCADE, blank=True, null=True
     )
-    sso_id = models.PositiveIntegerField(verbose_name='sso user.sso_id', default=None)
+    sso_id = models.PositiveIntegerField(verbose_name='sso user.sso_id', default=None, unique=False)
     export_countries = JSONField(blank=True, default=list)
     export_commodity_codes = JSONField(blank=True, default=list)
+    rules_regulations = JSONField(null=True, blank=True, default={})
```
richtier: `default=dict` is preferable as django sugest specifying callables rather than instances

PR #686, opened 2020-02-18
Feature/mvp add personalisation apis

WillTaylor22:
- [x] Changelog entry added.


conf/env/dev
```diff
@@ -16,7 +13,9 @@ DB_USER=debug
 DEBUG=true
 DEFAULT_FROM_EMAIL=debug
 DIRECTORY_CONSTANTS_URL_GREAT_DOMESTIC=http://profile.trade.great:8006/
-DIRECTORY_FORMS_API_BASE_URL=http://forms.trade.great:8011
+DIRECTORY_FORMS_API_API_KEY=tSPb18JExNbn2t3nXHuG9N2QHHxy2E9H671VSroENbW9DGnmvgGKkZZXbcKkpQQD
```
eadpearce: oh no!

personalisation/views.py
```diff
@@ -35,3 +38,115 @@ def create(self, request, *args, **kwargs):
...
+class EventsView(generics.GenericAPIView):
+    """ Events API - finds events near given geo-coordinates
+        URL parameters: 'sso_id' User's SSO_ID from cookie
+                        'lat'    Latitude
+                        'lng'    Longitude
+    """
+    permission_classes = []
+
+    def get(self, *args, **kwargs):
+        # sso_id = self.request.GET.get('sso_id', '')
```
eadpearce: remove commented code

personalisation/views.py
```diff
@@ -35,3 +38,115 @@ def create(self, request, *args, **kwargs):
...
+company’s creativity, expertise, technology and innovation offering.</p>
+''',
+                    "location": {
+                        "city": "London"
+                    },
+                    "url": "www.example.com"
+                },
+            ]}
+        )
+        # Actual
```
eadpearce: remove commented code

conf/env/dev
```diff
@@ -51,4 +50,15 @@ STANNP_TEST_MODE=true
...
+ACTIVITY_STREAM_INCOMING_ACCESS_KEY=some-id
+ACTIVITY_STREAM_INCOMING_SECRET_KEY=some-secret
+ACTIVITY_STREAM_OUTGOING_IP_WHITELIST='0.0.0.0, 0.0.0.0'
+ACTIVITY_STREAM_OUTGOING_ACCESS_KEY=incoming-some-id-1
+ACTIVITY_STREAM_OUTGOING_SECRET_KEY=incoming-some-secret-1
+ACTIVITY_STREAM_OUTGOING_URL=http://127.0.0.1:8080/v3/objects/_search
+EXPORTING_OPPORTUNITIES_API_BASIC_AUTH_USERNAME=debug
+EXPORTING_OPPORTUNITIES_API_BASIC_AUTH_PASSWORD=debug
+EXPORTING_OPPORTUNITIES_API_BASE_URL=http://exops.trade.great:3001
+EXPORTING_OPPORTUNITIES_API_SECRET=DkT49HM8TbOcsXL1L1jwDTEJALRYtt02fLiMAvNXzthpaHIMiFLw34YyfE1QpQ6bLc1d2tiLIZew7w1L5qwnXuQMjk5337aBiUmZeLuQyV2jqXxi8TsB9xiH6jKMFKVaY4P62mCp7rohsJhkQls8TBoYz37bfV1Qmt5Q4VZdOrWKSmpWtY3mGSE2AzMA1DJKy8FQ0FLW2RTe5NShLVZdB8cw2nU75peVopRjWB2J5c6M8j21unBnuJBs10
```
eadpearce: is this a 'live' secret? or is it only used in local

personalisation/helpers.py
```diff
@@ -0,0 +1,131 @@
...
+
+    if 'error' in content:
+        results = []
+        sentry_sdk.capture_message(
+            f"There was an error in /search: {content['error']}"
+        )
+    else:
+        results = serializers.parse_search_results(content)
+
+    return {'results': results}
```
eadpearce: could we avoid unnecessary nesting here by just returning `results`?

conf/env/dev
WillTaylor22: Doh. I have checked that the keys are changed and redeployed Forms dev and Dir API Develop. Also moved this key to the un-git-committed secrets file.

conf/env/dev
WillTaylor22: Double doh. I've changed this key & redeployed ExOps Dev + Profile Dev.

conf/env/dev
depsiatwal: we have all done this :-(

personalisation/helpers.py
WillTaylor22: I've added a comment to clarify.

In short, this is the hash of data returned from the API. When we want to add other data besides results (for example, 'page'), having a hash means we won't break the format of the API when adding additional data.

personalisation/views.py
WillTaylor22: Done.
Was in there as the removal is temporary, but you've rightly pointed out that the correct procedure is to store the commented code elsewhere.

personalisation/views.py
WillTaylor22: Done (as per comment above)

personalisation/helpers.py
eadpearce: ah i see. but couldn't we do that wherever we call that function instead of the function itself returning it?

PR #691, opened 2020-02-28
MVP-252 add cpi index

depsiatwal:
 - [x] Changelog entry added.


conf/urls.py
```diff
@@ -240,6 +240,11 @@
         dataservices.views.RetrieveEaseOfBusinessIndex.as_view(),
         name='dataservices-easeofdoingbusiness-index'
     ),
+    url(
+        r'^dataservices/corruptionperceptionsindex/(?P<country_code>.*)/$',
```
richtier: `corruption-perceptions-index` would be easier to read

dataservices/management/commands/import_cpi_data.py
```diff
@@ -0,0 +1,23 @@
...
+import tablib
+from import_export import resources
+from dataservices.models import CorruptionPerceptionsIndex
+
+
+class Command(BaseCommand):
+    help = 'Import CorruptionPerceptionsIndex data from transparency.org/'
+
+    def handle(self, *args, **options):
+        data = tablib.import_set(
```
richtier: to auto-close the file can we do:

```
with open('dataservices/resources/CPI2019.csv', 'r', encoding='utf-8-sig') as f:
        data = tablib.import_set(f.read(), format='csv', headers=True)
```

dataservices/management/commands/import_cpi_data.py
```diff
@@ -0,0 +1,23 @@
...
+            open('dataservices/resources/CPI2019.csv', 'r', encoding='utf-8-sig'),
+            format='csv',
+            headers=True
+        )
+        corruptionperceptionsindex_resource = resources.modelresource_factory(model=CorruptionPerceptionsIndex)()
+        result = corruptionperceptionsindex_resource.import_data(data, dry_run=True)
+        print(result.has_errors())
```
richtier: please use `self.stdout.write(` instead of `print`

dataservices/views.py
```diff
@@ -17,3 +17,17 @@ def get(self, *args, **kwargs):
...
+
+
+class RetrieveCorruptionPerceptionsIndex(RetrieveAPIView):
+    serializer_class = serializers.CorruptionPerceptionsIndexSerializer
+    permission_classes = []
+    lookup_url_kwarg = 'country_code'
+    lookup_field = 'country_code__iexact'
+    queryset = models.CorruptionPerceptionsIndex.objects.all()
+
+    def get(self, *args, **kwargs):
```
richtier: could simplify a bit with by using DRF's build-in exception handler method:
```

    def handle_exception(self, exc):
        if isinstance(exc, Http404):
            Response(data={})
        return super().handle_exception(exc)
```

PR #694, opened 2020-03-10
Mvp 276 comtrade data

depsiatwal:
 - [x] Changelog entry added.

### DIFF
--- a/dataservices/helpers.py
+++ b/dataservices/helpers.py
@@ -1,0 +1,75 @@
+import requests
+import json
+import pandas as pd
+import datetime
+
+
+class ComTradeData():
+    url = (
+        'https://comtrade.un.org/api/get?type=C&freq=A&px=HS&rg=1'
+    )
+
+    def __init__(self, commodity_code, reporting_area, partner_country='United Kingdom'):
+        pd.set_option('mode.chained_assignment', None)
+        self.product_code = self.get_product_code(commodity_code)
+        self.reporting_area_id = self.get_comtrade_company_id(reporting_area)
+        self.partner_country_id = self.get_comtrade_company_id(partner_country)
+
+    def get_comtrade_company_id(self, country_name):
+        with open('dataservices/resources/reporterAreas.json', 'r') as f:
+            json1_str = f.read()
+            counties_data = pd.DataFrame(json.loads(json1_str)['results'])
+            return counties_data[counties_data['text'] == country_name]['id'].iloc[0]
+
+    def get_url(self):
+        url_options = f'&r={self.reporting_area_id}&p={self.partner_country_id}&cc={self.product_code}&ps=All'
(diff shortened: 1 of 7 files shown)


dataservices/helpers.py
```diff
@@ -0,0 +1,80 @@
+import requests
+import json
+import pandas as pd
+import datetime
+
+
+class ComTradeData():
```
richtier: no need for paren

dataservices/helpers.py
```diff
@@ -0,0 +1,80 @@
...
+    url = (
+        'https://comtrade.un.org/api/get?type=C&freq=A&px=HS&rg=1'
```
richtier: can be single line

dataservices/helpers.py
```diff
@@ -0,0 +1,80 @@
+import requests
+import json
+import pandas as pd
```
richtier: why renaming as `pd`? bit harder to read. I thought "what's pd?" and had to scroll up to find out. If it was not renamed then devs will not have to do that :)

dataservices/helpers.py
```diff
@@ -0,0 +1,80 @@
...
+
+class ComTradeData():
+    url = (
+        'https://comtrade.un.org/api/get?type=C&freq=A&px=HS&rg=1'
+    )
+
+    def __init__(self, commodity_code, reporting_area, partner_country='United Kingdom'):
+        pd.set_option('mode.chained_assignment', None)
+        self.product_code = self.get_product_code(commodity_code)
+        self.reporting_area_id = self.__get_comtrade_company_id(reporting_area)
```
richtier: private methods make unit testing that method hard, and make reading a bit harder. It also makes devs wonder "why is this private". Overall net negative when it comes to maintainability.

dataservices/helpers.py
```diff
@@ -0,0 +1,80 @@
...
+        self.partner_country_id = self.__get_comtrade_company_id(partner_country)
+
+    def __get_comtrade_company_id(self, country_name):
+        with open('dataservices/resources/reporterAreas.json', 'r') as f:
+            json1_str = f.read()
+            counties_data = pd.DataFrame(json.loads(json1_str)['results'])
+            if not counties_data[counties_data['text'] == country_name]['id'].empty:
```
richtier: double negative. instead of `if not a == b` it's easier to read `if a != b`

richtier: what's happening here?

`counties_data[counties_data['text'] == country_name]`

that evaluates to `counties_data[True]` or `counties_data[False]`

depsiatwal: No this is used for filtering it actually returns a pandas dataseries of true/false values which is then used to filter the dataframe. 

Make sense ?

I could break up the line if it doesn't

PR #696, opened 2020-03-10
MVP-310 added views for comtrade data

depsiatwal:
 - [x] Changelog entry added.

### DIFF
--- a/dataservices/views.py
+++ b/dataservices/views.py
@@ -55,9 +55,7 @@ class RetrieveHistoricalImportDataView(generics.GenericAPIView):
 
         comtrade = helpers.ComTradeData(commodity_code=commodity_code, reporting_area=country)
 
-        historical_data = {'historical_import_data': []}
-        historical_data['historical_import_data'].append(comtrade.get_historical_import_value_partner_country())
-        historical_data['historical_import_data'].append(comtrade.get_historical_import_value_world())
+        historical_data = comtrade.get_all_historical_import_value()
         return Response(
             status=status.HTTP_200_OK,
             data=historical_data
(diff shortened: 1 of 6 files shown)


dataservices/views.py
```diff
@@ -29,3 +30,49 @@ def handle_exception(self, exc):
...
+        country = self.request.GET.get('country', '')
+
+        try:
+            comtrade = helpers.ComTradeData(commodity_code=commodity_code, reporting_area=country)
+            last_year_data = comtrade.get_last_year_import_data()
+            return Response(
+                status=status.HTTP_200_OK,
+                data={'last_year_data': last_year_data}
+            )
+        except HTTPError:
```
richtier: HttpError will not catch Timout, ssl and other request exceptions. Can solve by doing `except request.exceptions.RequestException`, which all request exceptions inherit from

dataservices/views.py
```diff
@@ -29,3 +30,49 @@ def handle_exception(self, exc):
...
+        commodity_code = self.request.GET.get('commodity_code', '')
+        country = self.request.GET.get('country', '')
+
+        try:
+            comtrade = helpers.ComTradeData(commodity_code=commodity_code, reporting_area=country)
+
+            historical_data = {'historical_import_data': []}
+            historical_data['historical_import_data'].append(comtrade.get_historical_import_value_partner_country())
+            historical_data[
+                'historical_import_data'].append(comtrade.get_historical_import_value_world())
```
richtier: looks like this can be one line

richtier: can these lines be something like 
```
country_data = comtrade.get_historical_import_value_partner_country()
world_data = comtrade.get_historical_import_value_world()
historical_data = {'historical_import_data': country_data + world_data}
```
bit easier to read

richtier: or better yet can we add a method to `ComTradeData` called `get_all_historical_import_value()` which does that for us so then we can do 

```
historical_data = {'historical_import_data': comtrade.get_all_historical_import_value()}
```

PR #699, opened 2020-03-19
Allow actions objectives update

depsiatwal:
 - [x] Changelog entry added.

### DIFF
--- a/exportplan/serializers.py
+++ b/exportplan/serializers.py
@@ -13,19 +13,11 @@ class CompanyObjectivesSerializer(serializers.ModelSerializer):
             'start_date',
             'end_date',
             'companyexportplan',
-            'pk',
         )
         extra_kwargs = {
             # passed in by CompanyExportPlanSerializer created/updated
             'companyexportplan': {'required': False},
         }
-
-    def to_internal_value(self, data):
-        if data.get('pk'):
-            # Attempting to make id available so we can ceck for new or exisiting
-            # when doing an update
-            data['id'] = data['pk']
-        return super().to_internal_value(data)
 
 
 class ExportPlanActionsSerializer(serializers.ModelSerializer):
@@ -38,7 +30,6 @@ class ExportPlanActionsSerializer(serializers.ModelSerializer):
             'is_reminders_on',
             'action_type',
             'companyexportplan',
-            'pk',
         )
         extra_kwargs = {
             # passed in by CompanyExportPlanSerializer created/updated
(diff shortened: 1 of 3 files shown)


exportplan/serializers.py
```diff
@@ -74,15 +77,32 @@ def create(self, validated_data):
...
-            objective_serializer = CompanyObjectivesSerializer(data={**objective, 'companyexportplan': instance.pk})
-            objective_serializer.is_valid(raise_exception=True)
-            objective_serializer.save()
+    def update(self, instance, validated_data):
 
-        for action in actions:
+        objectives = {}
+        actions = {}
+        if validated_data.get('company_objectives'):
+            objectives = validated_data.pop('company_objectives')
```
richtier: can simplify lines 86...91 to be instead:
```
objectives = validated_data.pop('company_objectives', {})
actions = validated_data.pop('export_plan_actions', {})
```

exportplan/serializers.py
```diff
@@ -74,15 +77,32 @@ def create(self, validated_data):
...
-            action_serializer = ExportPlanActionsSerializer(data={**action, 'companyexportplan': instance.pk})
-            action_serializer.is_valid(raise_exception=True)
-            action_serializer.save()
+        super().update(instance, validated_data)
+        self.recreate_objectives(instance, objectives)
+        self.recreate_actions(instance, actions)
         return instance
+
+    def recreate_objectives(self, instance, objectives):
+        instance.company_objectives.all().delete()
```
richtier: it's worth adding the `transation.atomic` decorator to `recreate_objectives` and `recreate_actions` to allow the `delete` to be rolled back if somehow the `create` failed

That avoids the problem that a report's objectives are deleted but creation failed, which would result in their objectives being lost

https://docs.djangoproject.com/en/3.0/topics/db/transactions/#django.db.transaction.atomic

exportplan/serializers.py
```diff
@@ -74,15 +77,32 @@ def create(self, validated_data):
...
+        for objective in objectives:
+            data = {**objective, 'companyexportplan': instance}
+            models.CompanyObjectives.objects.create(**data)
+
+    def recreate_actions(self, instance, actions):
+        instance.export_plan_actions.all().delete()
+        for action in actions:
+            data = {**action, 'companyexportplan': instance}
+            models.ExportPlanActions.objects.create(**data)
```
richtier: this will result in a database write per objective. could be quicker using `bulk_create`: https://docs.djangoproject.com/en/3.0/ref/models/querysets/#bulk-create

PR #700, opened 2020-03-23
Add recommended country end point

WillTaylor22:
 - [x] Changelog entry added.

Note that the failing test is unrelated to this PR


personalisation/management/commands/create_countries_of_interest.py
```diff
@@ -0,0 +1,23 @@
...
+            data = tablib.import_set(f.read(), format='csv', headers=True)
+
+            country_of_interest_resource = resources.modelresource_factory(model=CountryOfInterest)()
+
+            result = country_of_interest_resource.import_data(data, dry_run=True)
+            self.stdout.write(self.style.SUCCESS(result.has_errors()))
+            if not result.has_errors():
+                # No Errors lets flush table and import the data
+                CountryOfInterest.objects.all().delete()
+                country_of_interest_resource.import_data(data, dry_run=False)
```
richtier: can you add `@transaction.atomic` to `def_handle` please so the `CountryOfInterest.objects.all().delete()` is rolled back if `country_of_interest_resource.import_data` fails - thus avoiding situation where we delete interests but cannot import htem

personalisation/views.py
```diff
@@ -114,3 +115,16 @@ def get(self, *args, **kwargs):
...
                 data={'error_message': 'Connection to Export Opportunities failed'}
             )
+
+
+class RecommendedCountriesView(generics.ListAPIView):
+    serializer_class = serializers.CountryOfInterestSerializer
+    permission_classes = []
+
+    def get_queryset(self):
+        country = self.request.query_params.get('country', '').lower()
```
richtier: did not know about `query_params`. good to know :+1: 
https://www.django-rest-framework.org/api-guide/requests/#query_params

personalisation/views.py
```diff
@@ -114,3 +115,16 @@ def get(self, *args, **kwargs):
...
+        return models.CountryOfInterest.objects.\
+            filter(country=country).\
+            values('sector').\
+            annotate(num_sectors=Count('sector')).\
+            order_by('-num_sectors')[:5]
```
richtier: that's cool - ordering by annotated value, but throwing the calculated `num_sectors` away :+1:

personalisation/management/commands/create_countries_of_interest.py
WillTaylor22: Done

PR #702, opened 2020-03-25
MVP-363 - move MADB Airtable to backend

depsiatwal:
 - [x] Changelog entry added.


conf/settings.py
```diff
@@ -533,3 +533,5 @@
 EXPORTING_OPPORTUNITIES_API_BASIC_AUTH_PASSWORD = env.str('EXPORTING_OPPORTUNITIES_API_BASIC_AUTH_PASSWORD', '')
 EXPORTING_OPPORTUNITIES_API_BASE_URL = env.str('EXPORTING_OPPORTUNITIES_API_BASE_URL')
 EXPORTING_OPPORTUNITIES_API_SECRET = env.str('EXPORTING_OPPORTUNITIES_API_SECRET')
+# airtable
+AIRTABLE_API_KEY = env.str('AIRTABLE_API_KEY')
```
richtier: we're not going to be using this in prod, so this should be optional :)
```
 AIRTABLE_API_KEY = env.str('AIRTABLE_API_KEY', '')
```

PR #703, opened 2020-03-31
MVP359 - add multiple target-markets

depsiatwal:
 - [x] Changelog entry added.


exportplan/signals.py
```diff
@@ -0,0 +1,26 @@
+from dataservices import helpers
+
+
+def add_target_markets_data(sender, instance, *args, **kwargs):
+    existing_countries = []
+    if not instance._state.adding:
+        pre_save_instance = sender.objects.get(pk=instance.pk)
```
richtier: as we're only interested in `target_markets` we can do `pre_save_instance = sender.objects.only('target_markets').get(pk=instance.pk)` and that way we only read the fields from the db we care about. slightly more efficient

exportplan/signals.py
```diff
@@ -0,0 +1,26 @@
...
+        existing_countries = [market['country'] for market in pre_save_instance.target_markets]
+
+    for country in [market['country'] for market in instance.target_markets]:
```
richtier: this is looping twice - first to get the country then to loop over that country.

that's a bit harder to read compared with single loop:

```
    for target_market in instance.target_markets:
        country = target_market['country']

File: exportplan/signals.py [hunk 3]
```diff
@@ -0,0 +1,26 @@
...
+            instance.target_markets[index]['export_duty'] = rules_regulations['export_duty']
+            instance.target_markets[index]['easeofdoingbusiness'] = helpers.get_ease_of_business_index(
+                rules_regulations['country_code']
+            )
+            instance.target_markets[index]['corruption_perceptions_index'] = helpers.get_corruption_perception_index(
+                rules_regulations['country_code']
+            )
+            instance.target_markets[index]['last_year_data'] = helpers.get_last_year_import_data(
+                rules_regulations['country_code'], commodity_code
+            )
```
richtier: hard to read. this is bit easier:

```if country not in existing_countries:
    for target_market in instance.target_markets:
        if target_market["country"] == country
            break

    commodity_code = instance.export_commodity_codes[0]
    rules_regulations = helpers.MADB().get_rules_and_regulations(commodity_code)
    country_code = rules_regulations['country_code']
    target_market.update({
        'export_duty': rules_regulations['export_duty'],
        'easeofdoingbusiness': helpers.get_ease_of_business_index(country_code),
        'corruption_perceptions_index':  helpers.get_corruption_perception_index(country_code),
        'last_year_data': helpers.get_last_year_import_data(country_code, commodity_code),
    })
```

PR #710, opened 2020-04-16
MVP-398 add historical data feature flag

depsiatwal:
 - [x] Changelog entry added.


exportplan/signals.py
```diff
@@ -26,9 +27,13 @@ def add_target_markets_data(sender, instance, *args, **kwargs):
...
             'last_year_data': helpers.get_last_year_import_data(commodity_code=commodity_code, country=country),
-            'historical_import_data': helpers.get_historical_import_data(
-                commodity_code=commodity_code, country=country
-            ),
             'timezone': timezone,
             'utz_offset': datetime.now(pytz.timezone(timezone)).strftime('%z')
         })
+
+        if settings.FEATURE_COMTRADE_HISTORICAL_DATA_ENABLED:
+            target_market.update({
```
richtier, line 35: it's one key so no need to use `.update` - could be simpler:

```
historic_import_data = helpers.get_historical_import_data(commodity_code=commodity_code, country=country)
target_market['historical_import_data'] = historic_import_data
```

PR #713, opened 2020-05-04
Mvp 395 cache com trade data

depsiatwal:
 - [x] Changelog entry added.


dataservices/helpers.py
```diff
@@ -39,7 +41,7 @@ def get_last_year_import_data(self):
         comdata_df = pandas.DataFrame.from_dict(comdata.json()['dataset'])
         if not comdata_df.empty:
             # Get Last two years data
-            last_year = datetime.datetime.today().year-2
+            last_year = datetime.today().year-2
```
richtier, line 44: `last_year` vs `previous_year`. Normally I think those are synonyms. Naming is hard :)

dataservices/helpers.py
```diff
@@ -112,6 +114,31 @@ def get_rules_and_regulations(self, country):
             return rules[0]['fields']
 
 
+class cached(object):
```
richtier, line 117: no need to use `object` as a superclass any more :)

`cached` what? I think `TTLCache` would be more apt.

dataservices/helpers.py
```diff
@@ -112,6 +114,31 @@ def get_rules_and_regulations(self, country):
...
+
+    def get_cache_value(self, key):
+        return cache.get(key, default=None)
+
+    def set_cache_value(self, key, value):
+        cache.set(key, value, self.default_max_age)
+
+    def __call__(self, func):
+        def inner(*args, **kwargs):
+            cache_key = func.__name__ + ':' + '_'.join(list(args))
```
richtier, line 131: worth considering kwargs too

dataservices/helpers.py
```diff
@@ -112,6 +114,31 @@ def get_rules_and_regulations(self, country):
...
+            cached_value = self.get_cache_value(cache_key)
+            if not cached_value:
+                res = func(*args, **kwargs)
+                cached_value = {'data': res, }
```
richtier, line 135: why nest the data? could it be `self.set_cache_value(cache_key, res)`?

dataservices/helpers.py
depsiatwal: possibly I think it works for now , lets see how it evolves as I think data services we support will change drastically :-)

richtier: it will have unexpected outcomes when kwargs are used.

let's say some dev in future that is not either of us (or is us and we're having a slow day) changes this:

```
get_ease_of_business_index(user_id, country_code)
```

then later `get_ease_of_business_index(user_id=1, country_code='en-gb')` happens. Now all users will get user 1's because the caching key ignores kwargs.

also if someone refactors the calls and changes to call with kwarg like so: `get_ease_of_business_index(country_code='en-gb')` - then all countries will get the same result.

I think this is a key change we need to make :)

PR #714, opened 2020-05-06
Personalise events by user sector and country of interest

richtier:

 - [x] Changelog entry added.
 - [x] (if there are vulnerable requirements) Upgraded any vulnerable dependencies.
 
events-expertise

### DIFF
--- a/personalisation/helpers.py
+++ b/personalisation/helpers.py
@@ -10,28 +10,47 @@ from personalisation import serializers
 
 
 def parse_results(response):
-    content = json.loads(response.content)
+    content = response.json()
     # Hash of data & metadata (e.g. number of results) to return from API
     # Currently only provides 'results' but scope to expand
     return {'results': serializers.parse_search_results(content)}
 
 
-def build_query(lat, lng):
+def build_query(lat, lon, terms):
+    should = [
+        {
+            'multi_match': {
+                'query': term,
+                'fields': ['content', 'name'],
+                'fuzziness': 'AUTO',
+            }
+
+        } for term in terms
+    ]
+    # events closer to the user's location will be ranked higher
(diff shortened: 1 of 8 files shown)


personalisation/helpers.py
```diff
@@ -10,28 +10,47 @@
...
+            }
+
+        } for term in terms
+    ]
+    # events closer to the user's location will be ranked higher
     return json.dumps({
         'query': {
-          'match_all': {}
-        },
-        'sort': [{
```
richtier, line 24: had to replace the existing sorting because otherwise the search terms would not be respected - it would find documents by the provided terms but then order the results by how close the venue is to the user, meaning a result with a relevance score of 0.001 would be ahead of one with relevance score of 0.999 if the 0.001 one was geographically closer.

Replaced that sorting with [exponential decay function](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html) - so values close to the user are boosted, but the relevance score is still respected,

personalisation/views.py
```diff
@@ -55,40 +55,24 @@ def get_location(self):
...
-                    status=response.status_code,
-                    data=helpers.parse_results(response)
-                )
+        query = helpers.build_query(
+            lat=lat,
+            lon=lon,
+            terms=self.get_search_terms()
+        )
+        response = helpers.search_with_activitystream(query)
+        response.raise_for_status()
```
richtier, line 74: this has more-or-less the same outcome as before: a non-200 response is returned and the error ends up in sentry.

personalisation/helpers.py
WillTaylor22: Awesome!

PR #715, opened 2020-05-12
allow brand product details

depsiatwal:
 - [x] Changelog entry added.


exportplan/models.py
```diff
@@ -16,6 +16,7 @@ class CompanyExportPlan(TimeStampedModel):
     export_countries = JSONField(blank=True, default=list)
     export_commodity_codes = JSONField(blank=True, default=list)
     rules_regulations = JSONField(null=True, blank=True, default=list)
+    brand_product_details = JSONField(null=True, blank=True, default=list)
```
richtier, line 19: based on the factory I think this should be `default=dict`

PR #717, opened 2020-05-27
MVP-474 prepopulate-comtrade-cache

depsiatwal:
 - [x] Changelog entry added.


dataservices/management/commands/load_cache_seed_data_comtrade.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Command(BaseCommand):
+    help = 'Preload seed data for dataservice comtrade cache'
+
+    def handle(self, *args, **options):
+        countries = helpers.MADB().get_madb_country_list()
+        commodity_code = '2208.50.00.57'
+        class_name = 'dataservices.helpers'
+        countries = helpers.MADB().get_madb_country_list()
```
sdonk, line 12: This is the same as line 9

dataservices/management/commands/load_cache_seed_data_comtrade.py
```diff
@@ -0,0 +1,25 @@
...
+        function_names = ['get_last_year_import_data', 'get_last_year_import_data']
```
sdonk, line 13: same function?

richtier:
Unless I've missed something important the two commands can be simplified into one command:

```
class Command(BaseCommand):
    help = 'Populates the dataservice caches'
    def handle(self, *args, **options):
        for country in helpers.MADB().get_madb_country_list():
            kwargs = {'commodity_code': '2208.50.00.57', 'country': country[0]}
            
            dataservices.helpers.get_last_year_import_data(**kwargs)
            dataservices.helpers.x(**kwargs)
            dataservices.helpers.y(**kwargs)
            dataservices.helpers.z(**kwargs)
```

the current approach of:

- developer hard-codes the name of the function in `load_cache_seed_data_comtrade.py`
- developer calls load_cache_seed_data_comtrade in dev/staging/prod resulting in strings being added to the database
- populate_dataservices_cache is called in prod periodically, resulting in the function names being pulled from the database and called

could be simplified to:
- the_command_I_write_above is called in prod periodically, resulting in the function being called

no database hits, not manual labour, easy to understand

PR #719, opened 2020-06-04
mvp-581 added new fields

depsiatwal:
 - [x] Changelog entry added.


CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ MVP-416 - Personalise events by user sector and country of interest
 MVP-395 - cache comtrade data using backend cache
 MVP-479 - Brand Product Details
 MVP-474 - pre-populate dataservices cache task
+MVP-581 - new objective fields
```
richtier, line 40: added to the wrong place. please add to `Pre release` section

exportplan/models.py
```diff
@@ -36,9 +36,11 @@ class CompanyExportPlan(TimeStampedModel):
 
 class CompanyObjectives(TimeStampedModel):
     description = models.TextField(null=True, blank=True, default='', validators=[no_html])
-    owner = models.PositiveIntegerField(null=True, verbose_name='sso user.sso_id', default=None, unique=False)
+    planned_reviews = models.TextField(null=True, blank=True, default='', validators=[no_html])
```
richtier, line 39: if default is set to some value then no need to do `null=True` - as an empty value will be saved as `''` (unless explicitly set to `None`.

exportplan/models.py
```diff
@@ -36,9 +36,11 @@ class CompanyExportPlan(TimeStampedModel):
...
+    owner = models.TextField(null=True, blank=True, max_length=100)
```
richtier, line 40: that value would go in owner? I see it used to be `models.PositiveIntegerField(null=True, verbose_name='sso user.sso_id', default=None, unique=False)` so wondering why changed :)

depsiatwal: Ahh this is the owner against objective , they changed requirements as they want it as free text now :-(

PR #720, opened 2020-06-05
MVP-432 fixed adding countries error

depsiatwal:
 - [x] Changelog entry added.


dataservices/helpers.py
```diff
@@ -25,7 +25,7 @@ def get_comtrade_company_id(self, country_name):
             if not counties_data[counties_data['text'] == country_name]['id'].empty:
                 return counties_data[counties_data['text'] == country_name]['id'].iloc[0]
             else:
-                return ''
+                raise NotFound(f'Reporting country not found : {country_name}')
```
richtier, line 28: this error is django rest framework specific. I would expect the view to be raising this error, not a helper function :) It's giving the helper function functionality that belongs in the view

you could raise a custom error here and catch it in the view, so here `raise NoCountryFound()`

I imagine NoCountryFound would be a subclass of `NotImplementedError`

doing it this way has the advantage that the helper function is not tied to a view: The function will be agnostic of what calls it

PR #723, opened 2020-06-16
Gp2 154 update country list

depsiatwal:
 - [x] Changelog entry added.


exportplan/helpers.py
```diff
@@ -1,5 +1,10 @@
 import pytz
-from iso3166 import countries_by_alpha3
+from iso3166 import countries_by_alpha3, countries_by_name
+
+
+def get_iso3_by_country_name(country_name):
+    if countries_by_name.get(country_name.upper()):
```
richtier, line 6: is `country_name` guaranteed to be the name iso3166 expects? for reference elsewhere we handle that case by doing 

```
from difflib import SequenceMatcher

@functools.lru_cache(maxsize=None)
def is_fuzzy_match(label_a, label_b):
    match = SequenceMatcher(None, label_a, label_b)
    return match.ratio() > 0.9
```

could use that by looping over countries y name and checking each against the `country_name`. No action needed, just an FYI

exportplan/signals.py
```diff
@@ -16,23 +16,34 @@ def add_target_markets_data(sender, instance, *args, **kwargs):
...
             continue
+
+        country_code = export_helpers.get_iso3_by_country_name(country)
+
+        commodity_code = instance.export_commodity_codes[0]
+        timezone = export_helpers.get_timezone(country_code)
+
+        target_market.update({
+            'easeofdoingbusiness': helpers.get_ease_of_business_index(
+                country_code),
```
richtier, line 27: could be one line

exportplan/signals.py
```diff
@@ -16,23 +16,34 @@ def add_target_markets_data(sender, instance, *args, **kwargs):
         country = target_market['country']
         if country in pre_save_target_markets:
             continue
+
+        country_code = export_helpers.get_iso3_by_country_name(country)
+
+        commodity_code = instance.export_commodity_codes[0]
+        timezone = export_helpers.get_timezone(country_code)
```
richtier, line 23: `get_iso3_by_country_name` can return `None`. Does `get_timezone` handle `None`?

exportplan/tests/test_helpers.py
```diff
@@ -15,3 +15,7 @@ def test_get_timezone():
 
 def test_get_local_time_not_found():
     assert helpers.get_timezone('XS') is None
+
+
+def test_get_iso3_by_country_name():
+    assert helpers.get_iso3_by_country_name('Australia') == 'AUS'
```
richtier, line 21: please add tests for unhappy path and edge cases: passing in lowercase, uppercase, None, '', and a value that will not match against any country

exportplan/helpers.py
depsiatwal: Currently this is uses a gov list which would have been vetted for ISO codes they include 2 letter ISO codes. At some point we will move to use the 2 letter ISO.

PR #725, opened 2020-06-18
GP2-168 add factbook data

depsiatwal:
 - [x] Changelog entry added.


dataservices/models.py
```diff
@@ -22,3 +23,16 @@ class CorruptionPerceptionsIndex(TimeStampedModel):
 
     def __str__(self):
         return self.country_name
+
+
+class CIA_Factbook(TimeStampedModel):
```
richtier, line 28: CamelCase pleas e.g, CIAFactbook or CiaFactbook etc

PR #728, opened 2020-06-25
Add hs_code to models

WillTaylor22:
- [x] Changelog entry added.


supplier/migrations/0003_auto_20180109_0851.py
```diff
@@ -35,8 +35,4 @@ class Migration(migrations.Migration):
         ('supplier', '0002_auto_20180103_1159'),
     ]
 
-    operations = [
-        migrations.RunPython(
-            reset_supplier_primary_key_sequence, migrations.RunPython.noop
-        )
-    ]
+    operations = []
```
webbyfox, line 38: Not sure, why this been removed? fix migration although this will not resync PK as intended

webbyfox:
@depsiatwal I've added unit tests and further edit to align with the existing library.

PR #729, opened 2020-07-01
GP2-169 -load-IMF-Data

depsiatwal:
 - [x] Changelog entry added.


dataservices/tests/test_views.py
```diff
@@ -91,6 +125,42 @@ def test_get_corruptionperceptionsindex_not_found(api_client):
...
+
+@pytest.mark.django_db
+def test_get_worldeconomicoutlook_not_found(api_client):
+
+    url = reverse(
+        'dataservices-world-economic-outlook', kwargs={'country_code': 'xxx'}
+    )
+
+    response = api_client.get(url)
+    assert response.status_code == 200
```
webbyfox, line 160: Should API resulting in 404 for a country not exist

dataservices/admin.py
```diff
@@ -62,6 +62,30 @@ class CorruptionPerceptionsIndexAdmin(admin.ModelAdmin):
...
+class WorldEconomicOutlookAdmin(admin.ModelAdmin):
+    formfield_overrides = {
+        TextField: {'widget': forms.TextInput}
+    }
+
+    search_fields = (
+        'country_name',
+        'country_code',
+        'subject',
+        'scale'
```
webbyfox, line 75: comma at the end

dataservices/management/commands/import_weo_data.py
```diff
@@ -0,0 +1,22 @@
+from django.core.management import BaseCommand
+import tablib
+from import_export import resources
```
webbyfox, line 3: Space before to segregate project import from third-party/django import

dataservices/tests/test_views.py
depsiatwal: not for data services if we don't retrieve a company means we have no data and won't display anything. 
best effort basis we have too many different data sources if we raise exceptions we will need to handle them. Don't think it's required as it's optional

PR #731, opened 2020-07-02
migration fix for new build

webbyfox:
Migration was breaking for a new build. One of migration file doing some data massaging and expect record exists on `supplier` table which is not the case for a new build.


webbyfox:
OKay, updated CL

PR #737, opened 2020-07-13
Gp2 187 pull views factbook

depsiatwal:
 - [x] Changelog entry added.


dataservices/tests/factories.py
```diff
@@ -0,0 +1,14 @@
...
+import factory.fuzzy
+
+from dataservices import models
+
+
+class CIAFactBookFactory(factory.django.DjangoModelFactory):
+
+    country_key = 'united_kingdom'
+    country_name = 'United Kingdom'
+    factbook_data = {'population': '60m', 'capital': 'London', 'currency': 'GBP'}
```
webbyfox, line 11: This factory will always create the same object every time it been called. It might be ok for now but would be nice to introduce dynamic values to make use of it to other tests

depsiatwal: yeah we have a few tests of this nature where we are relying on static data. 
I agree we should try to move to a model where our factory can randomise the data for testing. I know some of the integrations tests make use of this.

PR #744, opened 2020-07-28
GP2-188 add total target age population

depsiatwal:

### DIFF
--- a/dataservices/helpers.py
+++ b/dataservices/helpers.py
@@ -134,6 +134,11 @@ class PopulationData:
                 population_data['rural_population_total']/population_data['total_population'], 6
             )
 
+        if population_data.get('male_target_age_population') and population_data.get('female_target_age_population'):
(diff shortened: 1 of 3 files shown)


dataservices/helpers.py
```diff
@@ -134,6 +134,11 @@ def get_population_data(self, country, target_ages):
                 population_data['rural_population_total']/population_data['total_population'], 6
             )
 
+        if population_data.get('male_target_age_population') and population_data.get('female_target_age_population'):
+            population_data['total_target_age_population'] = (
```
webbyfox, line 138: Can this not get by `get_population_total_data` https://github.com/uktrade/directory-api/blob/9bfa6faec7a45983089b23d032ed18edaa84195a/dataservices/helpers.py#L175

webbyfox: OK I see its age_targeted total

PR #745, opened 2020-08-03
UN data corrections

depsiatwal:
 - [x] Changelog entry added.


dataservices/tests/test_views.py
```diff
@@ -232,9 +232,9 @@ def test_get_cia_factbook_data(api_client):
     response = api_client.get(url, data={'country': 'United Kingdom', 'data_key': 'people, languages'})
 
     assert response.status_code == 200
+    print(response.json())
```
webbyfox, line 235: `print`?

PR #747, opened 2020-08-04
GP2-315 - add route to markets

depsiatwal:
 - [x] Changelog entry added.


exportplan/tests/factories.py
```diff
@@ -42,6 +42,17 @@ class Meta:
         model = models.CompanyObjectives
 
 
+class RouteToMarketsFactory(factory.django.DjangoModelFactory):
+
+    route = 'This is main route'
+    promote = 'Online marketing'
+    market_promotional_channel = 'Direct sales'
```
webbyfox, line 49: Can we make the above three fields dynamic, please?

PR #758, opened 2020-09-28
GP2-543-report-source-data-against-country-name 

depsiatwal:
 - [x] Changelog entry added.


depsiatwal:
Probably not the most elegant solution but mapping errors were too small and this works :-)

dataservices/helpers.py
```diff
@@ -93,6 +93,23 @@ class PopulationData:
     un_rural_pop = None
     un_urban_pop = None
     year = 0
+    country_map = {
```
stevejalim, line 96: Nit/Tiny: what's the purpose of this country map? Maybe a more descriptive var name would help?

dataservices/helpers.py
```diff
@@ -102,6 +119,19 @@ def __init__(self):
...
+        # Lets make the counties columns consistent with out own source
+        self.un_male_pop_data = self.map_country_data(self.un_male_pop_data)
+        self.un_female_pop_data = self.map_country_data(self.un_female_pop_data)
+        self.un_rural_pop = self.map_country_data(self.un_rural_pop)
+        self.un_urban_pop = self.map_country_data(self.un_urban_pop)
+
+    def map_country_data(self, country_data):
+        # This function is used to map the DS to match our naming convention for countries
+        country_data.country_name = country_data.country_name.apply(
+            lambda x: self.country_map.get(x) if self.country_map.get(x) is not None else x
```
stevejalim, line 131: Can we get away with just this, using `x` as the default val on a key miss?

```suggestion
            lambda x: self.country_map.get(x, x)
```

dataservices/views.py
```diff
@@ -76,10 +76,28 @@ def get(self, *args, **kwargs):
...
+        'St Kitts and Nevis': 'St. Kitts and Nevis',
+        'St Lucia': 'St. Lucia',
+        'St Vincent': 'St. Vincent and the Grenadines',
+        'Russia': 'Russian Federation',
+        'Syria': 'Syrian Arab Republic',
+        'The Bahamas': 'Bahamas, The',
+        'The Gambia': 'Gambia, The',
+        'Yemen': 'Yemen, Rep.',
+        'Venezuela': 'Venezuela, RB',
+    }
```
stevejalim, line 96: Should this dict match the one earlier? It doesn't at the mo

dataservices/views.py
```diff
@@ -101,6 +119,11 @@ def get(self, *args, **kwargs):
             data={'country_data': country_data}
         )
 
+    def map_country_data(self, country):
+        return (
+            self.country_map.get(country) if self.country_map.get(country) is not None else country
```
stevejalim, line 124: Again, can we use this alternative?

```suggestion
            self.country_map.get(country, country)
```

stevejalim (commented):
Overall seems fine but a couple of questions

stevejalim (approved):
Looks fine to me - Am assuming `get_population_total_data` is something we're making more robust in this changeset.

I get that `test_report_country_population_data_sets` is useful as a smoke test, but would be neat to have it more deeply tested if we're gonna rely on the report

PR #771, opened 2020-10-27
GP2-872 Reverse suggested markets order

bobmeredith:
 - [x] Changelog entry added.


webbyfox (approved):
LGTM

PR #774, opened 2020-10-29
Add companies endpoint for activity stream

niross:
New hawk authenticated endpoint at /activity-stream/companies

 - [x] Changelog entry added.


agiamas (approved):
lgtm

PR #775, opened 2020-11-02
Added delete endpoint for Company, CompanyUser

webbyfox:
 - [x] Changelog entry added.


company/permissions.py
```diff
@@ -1,8 +1,20 @@
...
+
+
+class ValidateDeleteRequest(permissions.BasePermission):
+    """Allow token access to data science team."""
+
+    def has_permission(self, request, view):
+        try:
+            return request.parser_context['kwargs']['request_key'] == settings.SSO_SIGNATURE_SECRET
+        except: # noqa
+            return False
```
depsiatwal, line 20: any reason to swallow all exception ?

company/views.py
```diff
@@ -29,6 +30,53 @@ def get(self, request, *args, **kwargs):
...
+    def delete(self, request, *args, **kwargs):
+        """
+        delete endpoint will take sso_id (user_id) as kwargs to delete company
+           > IF user is only user for associated company
+             then we delete user and company
+           > IF multiple users associated for a company and requested user is admin
+             then we re-assign admin role to other users
+             and delete user only, no company get deleted in this case.
+        """
+        sso_id = kwargs['sso_id']
```
depsiatwal, line 51: is the user not logged in at this point may be saver to use same pattern we use else where 
https://github.com/uktrade/directory-api/blob/develop/company/views.py#L42

company/views.py
```diff
@@ -29,6 +30,53 @@ def get(self, request, *args, **kwargs):
...
+        try:
+            request_user = models.CompanyUser.objects.get(sso_id=sso_id)
+        except models.CompanyUser.DoesNotExist:
+            return HttpResponse(status=204)
+
+        companies = models.Company.objects.filter(company_users__sso_id=sso_id)
```
depsiatwal, line 57: May be saver to use company associated to a user
https://github.com/uktrade/directory-api/blob/develop/company/views.py#L121

company/views.py
webbyfox: yes, we will be calling this as management command to archive `Company` and `CompanyUser` as data retention policy. This would be cron job eventually .

company/permissions.py
depsiatwal: Also should we not handle more cases here 
i.e Is the user an admin

webbyfox: Any exception raised, Code is restricting to use this endpoint via this permission class

depsiatwal (commented):
few comments inline

company/views.py
webbyfox: this would be called as a management command, there is no request user.

company/views.py
depsiatwal: ahh ok then do we need to expose via a rest/view 
may be restrict to execution via a management command only ?
That way no-one externally can access

depsiatwal (approved):
LGTM

PR #777, opened 2020-11-04
Added endpoint for target market data

webbyfox:
 - [x] Changelog entry added.

### DIFF
--- a/dataservices/tests/test_views.py
+++ b/dataservices/tests/test_views.py
@@ -112,7 +112,6 @@ def test_get_easeofdoingbusiness(api_client):
 
 @pytest.mark.django_db
 def test_get_easeofdoingbusiness_not_found(api_client):
-
     url = reverse(
         'dataservices-easeofdoingbusiness-index', kwargs={'country_code': 'xxx'}
     )
@@ -137,7 +136,6 @@ def test_get_corruptionperceptionsindex(api_client):
 
 @pytest.mark.django_db
 def test_get_corruptionperceptionsindex_not_found(api_client):
-
     url = reverse(
         'dataservices-corruptionperceptionsindex', kwargs={'country_code': 'xxx'}
     )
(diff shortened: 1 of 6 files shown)


dataservices/views.py
```diff
@@ -164,3 +164,35 @@ def get(self, *args, **kwargs):
             status=status.HTTP_200_OK,
             data={'population_data': population_data},
         )
+
+
+class RetrieveTargetMarketDataView(generics.GenericAPIView):
```
depsiatwal, line 169: Can we think of a more generic name here ?
Target markets is Magna / Export plans specific 

And dataservices is mean't to be a generic data services

dataservices/views.py
```diff
@@ -164,3 +164,35 @@ def get(self, *args, **kwargs):
...
+
+        countries = self.request.GET.getlist('country', '')
+
+        if not countries:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
+
+        data_set = []
+
+        for country in countries:
+            pop_data = helpers.PopulationData()
```
depsiatwal, line 182: whats difference between pop_data and population_data ?
may a different naming convention

depsiatwal: I know the difference :-) 

But can be confusing in the context 
maybe just call it population_data or country_population

dataservices/tests/test_views.py
```diff
@@ -383,3 +373,57 @@ def test_population_data_country_not_found(api_client):
...
+            'internet_usage': 90.97,
+            'rural_population_total': 10839,
+            'urban_population_total': 56495,
+            'total_population': 67888
+        }
+    ]
+
+
+@pytest.mark.django_db
+def test_target_market_data_view_multiple_countries(api_client, internet_usage_data):
```
depsiatwal, line 407: may want to change the test names also to reflect the views

depsiatwal (approved):
LGTM

PR #787, opened 2020-11-19
allow partial updates json fields

depsiatwal:

### DIFF
--- a/exportplan/serializers.py
+++ b/exportplan/serializers.py
@@ -2,6 +2,7 @@ from django.db import transaction
 from rest_framework import serializers
 
 from exportplan import models
+from django.contrib.postgres.fields import JSONField
 
 
 class CompanyObjectivesSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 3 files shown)


exportplan/serializers.py
```diff
@@ -146,9 +146,25 @@ def create(self, validated_data):
...
+        # This will allow partial updating to json fields during a patch update. Json fields generally represent
+        # a export plan page. we only want to update the field being sent else by nature we would wipe all the
+        # other fields
+        for field_name in validated_data.keys():
+            field_value = getattr(instance, field_name)
+            # Check if the field we are updating is JSON Type and ensure contents are JSON
+            if (
+                    isinstance(models.CompanyExportPlan._meta.get_field(field_name), JSONField) and
+                    isinstance(field_value, dict)
+            ):
```
webbyfox, line 162: don't you think we should raise the validation error if the above condition is not satisfied? also, worth adding negative test for this condition.

depsiatwal: This isn't field validation, it's more a conditional statement 
1) source is JSON
2) Incoming is JOSN 
if these conditions are met we do partial update based on incoming keys. 
Else we do a normal field update. 

This is tested by an additional test by updating non Json fields.

webbyfox: Ahh, I see.

webbyfox (approved):

PR #788, opened 2020-11-23
Added new endpoint for ComTrade import data from UK

webbyfox:

 - [x] Changelog entry added.


depsiatwal (approved):
LGTM

PR #790, opened 2020-11-25
Added GDP Per Capita data into directory-api

webbyfox:

 - [x] Changelog entry added.


rafa-garcia (approved):
lgtm

PR #791, opened 2020-11-27
Filter out companies from activity stream if date_published is null

abbas123456:
The company object returned from the activity stream endpoint has a published key which is set to the company's date_published. As this field can be null, unpublished companies need to be filtered out


agiamas (approved):
lgtm

PR #800, opened 2020-12-09
GP2-1068:  Apply Black and isort to the codebase + tooling to allow their ongoing use

stevejalim:
This massive changeset is the result of applying black and isort to the codebase.

I have visually scanned every changed file, in case it's too much for eyes-on CR here. Tests are passing

* add Black and related deps to test requirements and rebuild requirements
* add minimal Black config in pyproject.toml which will be used by default
* config for isort so that it doesn't conflict with Black
* add makefile commands to use black in check or write mode (`make checks` or `make autoformat`)
* update CI config to run `black --check` in CI, too
* add .pre-commit-config.yaml for pre-commit (see pre-commit.com)
* reformat the entire codebase (done as six separate commits to make review easier)

 - [X] Changelog entry added.
 - [X] (if updating requirements) Requirements have been compiled.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,6 +77,20 @@ jobs:
             . venv/bin/activate
             pip install flake8
             flake8
+  black:
+    docker:
+      - image: circleci/python:3.6.6
+    steps:
+      - checkout
+      - setup_remote_docker:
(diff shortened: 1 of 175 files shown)


webbyfox (approved):
LGTM

PR #810, opened 2021-01-14
Add endpoint for society data

rafa-garcia:
 - [X] Changelog entry added.

### DIFF
--- a/dataservices/helpers.py
+++ b/dataservices/helpers.py
@@ -280,6 +280,21 @@ class PopulationData:
         return urban_rural_data
 
 
+class SocietyData:
+    def get_people_data(self, country):
+        people_data = {}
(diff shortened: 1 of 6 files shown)


dataservices/helpers.py
```diff
@@ -280,6 +280,21 @@ def get_population_urban_rural_data(self, country, classification):
         return urban_rural_data
 
 
+class SocietyData:
+    def get_people_data(self, country):
```
webbyfox, line 284: Having just this helper method will do, don't think `SocietyData` helps anyway here and add `TTLCache` to the method as this won't change

webbyfox (changes requested):

webbyfox (approved):
LGTM

PR #816, opened 2021-01-22
GP2-1348 Economy and population - added raw values and rank_totals

bobmeredith:
This PR just adds some fields to api responses.
A raw, unformatted population field so that we can format correctly in the UI layer.
Totals for corruption perception and ease of business, Also hardcoded years for these and a 'rank' field that duplicates the 'year_2019' field.
This means that great-cms can be made independent of new data being loaded in directory-api.
Existing fields are unchanged so maintaining backward compatibility. 

 - [x] Changelog entry added.

### DIFF
--- a/dataservices/views.py
+++ b/dataservices/views.py
@@ -119,7 +119,6 @@ class RetrieveCountryDataView(generics.GenericAPIView):
     permission_classes = []
 
     def get(self, *args, **kwargs):
-
         country = self.map_dit_to_weo_country_data(self.kwargs['country'])
         filter_args = {'country_name': country}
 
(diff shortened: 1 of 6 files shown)


dataservices/views.py
```diff
@@ -237,7 +235,10 @@ def get(self, *args, **kwargs):
             country_population = helpers.PopulationData()
             country_data = {'country': country}
             total_population = country_population.get_population_total_data(country=country)
-            population_data = {'total_population': millify(total_population.get('total_population', 0) * 1000)}
+            population_data = {
+                'total_population': millify(total_population.get('total_population', 0) * 1000),
+                'total_population_raw': total_population.get('total_population', 0) * 1000,
```
webbyfox, line 240: Bob, Do you mind adding a new variable for raw value on top to avoid calculating twice?

webbyfox (approved):
Just minor, but good

dataservices/views.py
bobmeredith: Fair enough.

PR #824, opened 2021-02-01
Added Trading blocs to data services

webbyfox:
 - [x] Changelog entry added.


depsiatwal (commented):
For other CSV we used the resource loader lib to import.
Does this not work for this CSV , may be it's not flexible enough ?

webbyfox:
yeah, it not. like figuring out the foreign key for the country model and formatting data before adding

depsiatwal (approved):
LGTM

PR #830, opened 2021-02-05
[GP2-1392] Get resources from S3

rafa-garcia:
To do:

 - [X] Changelog entry added.


dataservices/management/commands/import_suggested_countries.py
```diff
@@ -1,38 +1,42 @@
...
-                            SuggestedCountry(hs_code=hs_code, country=country, order=order)
-                        )
-                        order += 1
-
-            SuggestedCountry.objects.all().delete()
-            SuggestedCountry.objects.bulk_create(suggested_product_countries)
+        file_name = 'SR.CREST.ALL.csv'
+        bucket_name = settings.AWS_STORAGE_BUCKET_NAME_DATA_SCIENCE
+        s3_resource = get_file_from_s3(bucket_name, file_name)
+        csv_file = s3_resource['Body'].read().decode('utf-8')
```
webbyfox, line 16: can we keep `with` as previously

bobmeredith: better yet, could you put all the code to get s3 file in a fn in helpers?
Maybe default the bucket name too.

`       def open_data_science_file(file_name,
            bucket_name=settings.AWS_STORAGE_BUCKET_NAME_DATA_SCIENCE):
            s3_resource = get_file_from_s3(bucket_name, file_name)
            return s3_resource['Body'].read().decode('utf-8')
`
then call as:
`
with open_data_science_file(file_name='SR.CREST.ALL.csv') as f:
    data = tablib.import_set(f, format='csv', headers=True)
    ....
`

rafa-garcia: Boto3 client gives us a S3 resource object, whose `Body` is a stream. It doesn't behave as `open()`, which opens a file and return a corresponding file object.

`with` requires an object with the necessary `__enter__()` and `__exit__()` methods. Is it worth it to write a class to handle the reading of the stream?

webbyfox (approved):

PR #833, opened 2021-02-11
EP change to dict

depsiatwal:
 - [x] Changelog entry added.


rafa-garcia:
@depsiatwal would this affect the client in any way?

depsiatwal:
> @depsiatwal would this affect the client in any way?

nah client should be good as we save dict values anyway but dict should have been default anyway

rafa-garcia (approved):
LGTM

PR #838, opened 2021-02-19
GP2-1398_comtrade_data_copy

bobmeredith:
Moves a copy of filtered comtrade data into the directory-api database.
The new request to get comtrade data supports multiple countries, single commodity code and does no processing, so all the data are returned for the FE to compute changes and trends.
The country list is always a list of iso2 codes.
There is also a new request for country data that allows multiple country's data to be requested in one go, again using iso3. The model names required can be passed in so this can become a generic request to get any data for a country.

A compacted comtrade data extract is in S3.  run `make manage import_comtrade_data` to load it.
This process can take an hour or more I'm afraid.

The extract is generated as follows:
Comtrade data are downloaded as vast csv files (about 6GB), one for each year. There is a management command 
`make manage import_comtrade_data -- --raw  <filename>`
That reads through such files and loads your db with only the data that are needed.
Then, a sql dump of the table to generate the single data file that gets uploaded to S3.

 - [x] Changelog entry added.


CHANGELOG.md
```diff
@@ -22,7 +23,7 @@
 - GP2-1382 - getting paid structure
 - GP2-1180 - travel bus BE
 - NOTICKET - Exportplan make dict default for JSON fields.
-- GP2-1181 - business risk
+- GP2-1181 - 1398business risk
```
depsiatwal, line 26: look like a typo here

depsiatwal (approved):
Small comment just on CL
nicely done :-)

dataservices/tests/conftest.py
```diff
@@ -16,3 +16,53 @@ def internet_usage_data():
...
+    )
+    models.ComtradeReport.objects.create(
+        country=countries['NL'],
+        country_iso3='NLD',
+        year=2020,
+        commodity_code='123456',
+        trade_value=91.97,
+    )
+    yield
+    models.InternetUsage.objects.all().delete()
```
webbyfox, line 46: I guess you need to delete all records from `ComtradeReport` here

dataservices/tests/conftest.py
```diff
@@ -16,3 +16,53 @@ def internet_usage_data():
...
+        year_2019=12,
+    )
+    models.EaseOfDoingBusiness.objects.create(
+        country=countries['NL'],
+        country_code='NL',
+        country_name='Netherlands',
+        year_2019=13,
+    )
+    yield
+    models.InternetUsage.objects.all().delete()
```
webbyfox, line 67: similarly here `EaseOfDoingBusiness`

webbyfox (changes requested):
minor comments

dataservices/tests/conftest.py
bobmeredith: Hehe - well spotted - thanks

CHANGELOG.md
bobmeredith: Thanks for the spot.

webbyfox (approved):
LGTM

PR #839, opened 2021-02-27
GP2-1441 CPI update

bobmeredith:
Update to CPI data.
The original was very limited as we only pulled a single year of data into our db, and exposed these with a year-specific fieldname.
This PR updates to holding all available years in the db although the accessor picks the latest for each country.
Also includes updates to the loaders for other data to populate the 'country' foreign keys in each loading management command.

 - [ ] Changelog entry added.


depsiatwal (approved):
LGTM

PR #843, opened 2021-03-03
Updated CPI data

webbyfox:

 - [x] Changelog entry added.


rafa-garcia (approved):
LGTM

bobmeredith (commented):
These data look a bit wrong.
We've gone from 264 countries to 89! maybe load up a  sequence rather than just the last year.

dataservices/management/commands/tests/test_import_data.py
```diff
@@ -15,7 +15,7 @@
         (models.EaseOfDoingBusiness, 'import_easeofdoingbusiness_data', 264, 1),
         (models.WorldEconomicOutlook, 'import_weo_data', 1552, 0),
         (models.InternetUsage, 'import_internet_usage_data', 264, 1),
-        (models.ConsumerPriceIndex, 'import_consumer_price_index_data', 264, 1),
```
bobmeredith, line 18: Not sure this is right.  There are far fewer data now.

dataservices/management/commands/import_consumer_price_index_data.py
```diff
@@ -17,14 +15,21 @@ def handle(self, *args, **options):
...
-            cursor.execute(
-                "update dataservices_consumerpriceindex as d \
-                set country_id=c.id \
-                from dataservices_country c where d.country_code=c.iso3;"
-            )
+            cpi_data = []
+
+            for item in data:
+                try:
+                    country = Country.objects.get(iso3=item[1])
```
bobmeredith, line 22: Suggest using a csv reader rather than this tablib - then you can access by the fieldname from the header, rather than index.

bobmeredith (changes requested):

bobmeredith (approved):
LGTM

PR #844, opened 2021-03-05
Add population target age groups to dataservices

rafa-garcia:
To do:

 - [X] Changelog entry added.


dataservices/management/commands/import_target_age_groups.py
```diff
@@ -0,0 +1,56 @@
...
+class Command(BaseCommand):
+    help = 'Import target age groups (male/female).'
+
+    def handle(self, *args, **options):
+        genders = ['male', 'female']
+        population_data = []
+
+        for gender in genders:
+            with open(f'dataservices/resources/world_population_medium_{gender}.csv', 'r', encoding='utf-8-sig') as f:
+                data = tablib.import_set(f.read(), format='csv', headers=True)
```
bobmeredith, line 16: Suggest you use a csv DictReader rather than tablib as it allows you to access data by fieldname rather than index.  Dictionary fieldnames come from the csv header row.

dataservices/management/commands/import_target_age_groups.py
```diff
@@ -0,0 +1,56 @@
...
+                            age_80_84=item[20] if item[20] and item[20] != '...' else None,
+                            age_85_89=item[21] if item[21] and item[21] != '...' else None,
+                            age_90_94=item[22] if item[22] and item[22] != '...' else None,
+                            age_95_99=item[23] if item[23] and item[23] != '...' else None,
+                            age_100_plus=item[24] if item[24] and item[24] != '...' else None,
+                        )
+                    )
+
+            PopulationData.objects.all().delete()
+            PopulationData.objects.bulk_create(population_data)
```
bobmeredith, line 54: Heh, I like this.  I'm gonna see if I can refactor the comtrade import by doing this in blocks.

dataservices/models.py
```diff
@@ -224,3 +224,39 @@ class Meta:
...
         return str(self.country_iso3) + ':' + str(self.commodity_code)
+
+
+class PopulationData(models.Model):
+    """
+    Each record represents data from one gender, for one particular place and year.
+    """
+
+    country = models.ForeignKey(
+        'dataservices.Country', verbose_name=_('Countries'), on_delete=models.SET_NULL, null=True
```
bobmeredith, line 235: This worries me a little - I think the country import mgmt command deletes all the country objects before importing them.  It would thus leave all models like this orphaned and this model has no other country code to reconnect with.
Possibly should be 'Protect' but I guess this should be reviewed across the full suite of models.
The country model should really have its iso code as the PK and link to that rather than some arbitrary id.

dataservices/models.py
```diff
@@ -224,3 +224,39 @@ class Meta:
...
+    age_70_74 = models.PositiveIntegerField(null=True, default=None, unique=False)
+    age_75_79 = models.PositiveIntegerField(null=True, default=None, unique=False)
+    age_80_84 = models.PositiveIntegerField(null=True, default=None, unique=False)
+    age_85_89 = models.PositiveIntegerField(null=True, default=None, unique=False)
+    age_90_94 = models.PositiveIntegerField(null=True, default=None, unique=False)
+    age_95_99 = models.PositiveIntegerField(null=True, default=None, unique=False)
+    age_100_plus = models.PositiveIntegerField(null=True, default=None, unique=False)
+
+    class Meta:
+        verbose_name = "Target age groups"
```
bobmeredith, line 262: There should be some unique constraint here - country,year,gender I guess.

dataservices/serializers.py
```diff
@@ -105,3 +105,32 @@ def get_year(self, obj):
...
+            '60-64': {'source': 'age_60_64'},
+            '65-69': {'source': 'age_65_69'},
+            '70-74': {'source': 'age_70_74'},
+            '75-59': {'source': 'age_75_79'},
+            '80-84': {'source': 'age_80_84'},
+            '85-89': {'source': 'age_85_89'},
+            '90-94': {'source': 'age_90_94'},
+            '95-99': {'source': 'age_95_99'},
+            '100+': {'source': 'age_100_plus'},
+        }
```
bobmeredith, line 136: Shouldn't you exclude all the 'age_xx_xx' fields, to stop the data from being serialized twice? Looks good though.

bobmeredith (commented):
Nicely done.
A few comments though.

bobmeredith (approved):
LGTM

PR #845, opened 2021-03-08
GP2-1611 Filtering on country data

bobmeredith:
Allows arbitrary django filters to be applied to the country_data endpoint as well as a list of countries. SQL injection threats and security is maintained as only models in the dataservices app can be accessed and filtration is done through a dango 'filter' so no raw sql access is involved.
To use this, the list of fields should be supplied as a json string, containing model name, filter and optionally, a flag to return data for only the latest year in the filtered dataset (which may be more than one row in certain models)
 
 - [x] Changelog entry added.


webbyfox (approved):
LGTM

PR #847, opened 2021-03-11
GP2-1724 Speed up population tab + cleanup

bobmeredith:
Adds a model and loader for population divided by urban/rural
Removes old comtrade endpoints and associated tests including the exportplan signal to populate EP with comtrade data on a change of market.

 - [x] Changelog entry added.


webbyfox (approved):
Happy to see this cleanup

PR #860, opened 2021-03-25
Data retention - amended delete endpoint so it returns statistics

webbyfox:
 - [x] Changelog entry added.


depsiatwal (approved):
LGTM

PR #863, opened 2021-03-29
GP2-1915 Align EP data snapshots

bobmeredith:
Main focus of this PR - adding country FK to CIA factbook data. I added a management command to do this - also, since factbook only has country names, I had to create a mapping table to translate certain country names on load.

 - [x] Changelog entry added.


webbyfox (approved):
LGTM

PR #868, opened 2021-04-01
GP2-1920 Update countries data

bobmeredith:
Update to the countries reference data file to version 5.35
This is needed to correct errors in the numeric iso codes in certain countries.

Since most of our other data tables are keyed to countries, replacing this requires that all the other data be re-populated - or at least re-linked.
To ease this - I have created a management command that runs the other management commands to import data. it does NOT run `import_countries`, nor `import_comtrade_data` as both of these take a long time to run.

Comtrade is by far the largest table - and the initial step of import_countries takes a long time to run as it nulls the foreign keys through a django cursor. Thus, I added a couple of new options to the `import_comtrade_data` command to allow the countries to be linked and unlinked.

So - the full update process goes like:
```
make manage import_comtrade_data -- --unlink_countries
make manage import_countries
make manage import_all
make manage import_comtrade_data -- --link_countries
```

If there is little or no comtrade data in the system, the comtrade steps can be skipped.

 - [x] Changelog entry added.


depsiatwal (approved):
LGTM

PR #870, opened 2021-04-06
GP2-1920_population_delete_strategy

bobmeredith:
No CL as this is a tiny fix to previous PR
The PR https://github.com/uktrade/directory-api/pull/868 will not update on live as the delete strategy on the country FK on population data is set to 'do nothing'. Since there is a constraint - the update_countries management command fails on dev.

This PR corrects to SET_NULL instead.
Also, I added the population data to admin.

 - [ ] Changelog entry added.


rafa-garcia (approved):
LGTM

PR #876, opened 2021-04-13
GP2-1920_ref_countries_zero_padding

bobmeredith:
This PR fixes an issue caused by an update to our reference country list from the data team.
The new version of the country list has the iso1 numeric code, zero padded to three digits. This means that it no longer joins to the numeric iso1 field in population data from the UN.
To fix, I've changed the two loaders for the population data to zero pad the field on country look-up.


webbyfox (approved):
CL is missing but I am anti-CL

PR #895, opened 2021-06-07
GBAU-866-ch-verify-error 

depsiatwal:

 - [ ] Changelog entry added.


bobmeredith (approved):
Happy to approve, but this PR doesn't look like it fixes anything.  The default URLs are exactly the same as the hardcoded ones right?
Does give scope for setting in environments though

PR #917, opened 2021-06-30
GP2-3071 Improve make secrets script

froddd:
Changed `init_secrets` Make script to `secrets` for consistency with other directory repos.

Makefile secret script now uses sed directly to create the new file to avoid using -i option, which ensures compatibility with both Mac and Linux platforms (and prevents creating extra secrets-do-not-commit-e file on Mac).

The secrets file is only created if it does not exist, otherwise user is prompted to delete if they require a fresh one.

CONTEXT: This should make onboarding and fresh setup slightly easier and more consistent when using the Docker setup from https://github.com/uktrade/great-cms

 - [ ] Changelog entry added.
 - [ ] ~~(if there are vulnerable requirements) Upgraded any vulnerable dependencies.~~
 - [ ] ~~(if updating requirements) Requirements have been compiled.~~
 - [ ] ~~(if adding env vars) Added any new environment variable to vault.~~
 - [ ] ~~(if adding feature flags) Cleaned up old flags~~
 - [ ] ~~(if hotfix) Has made PR into develop too~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@
 
 #### Configuration
 
-Secrets such as API keys and environment specific configurations are placed in `conf/env/secrets-do-not-commit` - a file that is not added to version control. To create a template secrets file with dummy values run `make init_secrets`.
+Secrets such as API keys and environment specific configurations are placed in `conf/env/secrets-do-not-commit` - a file that is not added to version control. To create a template secrets file with dummy values run `make secrets`.
 
 ### Commands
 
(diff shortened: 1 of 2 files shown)


stevejalim:
> and prevents creating extra secrets-do-not-commit-e file on Mac

🙌 🙌 🙌 🙌 🙌 🙌

depsiatwal (approved):
LGTM

PR #920, opened 2021-07-08
Relabelling data services models

webbyfox:

 - [x] Changelog entry added.


stevejalim (approved):
Def no migration needed?

webbyfox:
Added migrations

PR #924, opened 2021-07-27
GP2-2896 Split objective start/end dates into month/year

froddd:
Updated models to represent Company objectives start and end dates as month/year instead of date. Migration replicates existing data into new columns, while previous columns have been kept as is to avoid data loss -- not sure this is entirely desirable given the nature of the data? TBC!

Related great-cms changes: https://github.com/uktrade/great-cms/pull/1470

 - [x] Changelog entry added.


depsiatwal (approved):
LGTM

PR #928, opened 2021-08-12
GP2-3179 Spike

seijihg:
Report ss:

Screenshot 2021-08-12 at 10 21 08

Screenshot 2021-09-06 at 21 25 07

 - [ ] Changelog entry added.


exportplan/helpers.py
```diff
@@ -16,3 +16,42 @@ def get_timezone(country_code):
...
+    foreign_key_sections = [
+        "company_objectives",
+        "exportplan_downloads",
+        "route_to_markets",
+        "target_market_documents",
+        "funding_credit_options",
+        "business_trips",
+        "business_risks",
+    ]
+    print(plan.ui_progress)
```
depsiatwal, line 42: no print statements

exportplan/management/commands/report_export_plan.py
```diff
@@ -0,0 +1,49 @@
...
+                    self.stdout.write(self.style.SUCCESS("This EP has content."))
+
+            # No country or product selected then checks everything elses.
+            else:
+                if is_ep_plan_empty(plan):
+                    empty_ep_counter += 1
+                    self.stdout.write(self.style.WARNING("EP is empty"))
+                    self.stdout.write("---")
+                else:
+                    not_empty_ep_counter += 1
```
depsiatwal, line 43: need test case else we drop code cov

exportplan/helpers.py
```diff
@@ -16,3 +16,42 @@ def get_timezone(country_code):
     iso3_country_code = country_code_iso3_to_iso2(country_code)
     if iso3_country_code and pytz.country_timezones(iso3_country_code):
         return pytz.country_timezones(iso3_country_code)[0]
+
+
+def is_ep_plan_empty(plan):
```
depsiatwal, line 21: don't think we need this as part of export plan since it has no use in prod code. please move to management helper class.

depsiatwal: can this not be read for the actual model using attributes or some fields collection ?

exportplan/management/commands/report_export_plan.py
```diff
@@ -0,0 +1,49 @@
...
+
+        for plan in export_plans.iterator():
+            self.stdout.write(self.style.SUCCESS(f'{plan.company}:')) if plan.company else self.stdout.write(
+                self.style.WARNING("No Company is associated with this EP.")
+            )
+
+            # With country or commodity checks.
+            if plan.export_commodity_codes or plan.export_countries:
+                self.stdout.write(
+                    self.style.SUCCESS(f'Picked Product: {plan.export_commodity_codes[0]["commodity_name"]}')
```
depsiatwal, line 22: why export_countries would for product check ?

exportplan/management/commands/report_export_plan.py
```diff
@@ -0,0 +1,49 @@
...
+                    empty_ep_counter += 1
+                    self.stdout.write(self.style.WARNING("EP is empty"))
+                    self.stdout.write("---")
+                else:
+                    not_empty_ep_counter += 1
+                    self.stdout.write(self.style.SUCCESS("This EP has content."))
+
+            # No country or product selected then checks everything elses.
+            else:
+                if is_ep_plan_empty(plan):
```
depsiatwal, line 38: can probably structure this better why check twice i.e 
else fall back

exportplan/management/commands/tests/test_report_export_plan.py
```diff
@@ -0,0 +1,107 @@
+import pytest
+from django.core import management
+from exportplan.models import CompanyExportPlan
+import json
+
+
+@pytest.fixture()
+def new_export_plan_factory(db):
```
depsiatwal, line 8: we already have a factory class is exportplan/tests where you can customise the creation

exportplan/management/commands/tests/test_report_export_plan.py
```diff
@@ -0,0 +1,107 @@
...
+
+@pytest.fixture
+def exportplan_A(db, new_export_plan_factory):
+    return new_export_plan_factory(
+        sso_id=1, export_countries=[{"region": "Africa", "country_name": "Angola", "country_iso2_code": "AO"}]
+    )
+
+
+@pytest.fixture
+def exportplan_B(db, new_export_plan_factory):
```
depsiatwal, line 52: fixture will have more meaning if we explicit name what they're testing. 
i.e export_plan_empty, export_plan_with_countries etc etc

depsiatwal (commented):
comments added
I think it needs more tests and a tidy up of the code flow. 

see comments for more detail.

exportplan/management/commands/report_export_plan.py
seijihg: its on the ticket itself where Alex asked for it?
Screenshot 2021-09-06 at 21 17 30

exportplan/management/commands/report_export_plan.py
seijihg: yes I am working on it with the newest PR :D

exportplan/helpers.py
seijihg: yes keeps forgetting these :D

exportplan/helpers.py
seijihg: tried to pull fields from models in the newest PR hope its good.

exportplan/management/commands/report_export_plan.py
seijihg: not sure I understand this one is checking when it has country/commodity then if FALSE then it goes to check without coutry/commodity  so it  can print out information in terminal as requested in the ticket.

exportplan/management/commands/tests/test_report_export_plan.py
seijihg: yes implemented! great learning there.

exportplan/management/commands/tests/test_report_export_plan.py
seijihg: done as well :D

exportplan/management/commands/report_export_plan.py
depsiatwal: will this not break
what happens if a user has plan.export_countries but hasn't selected  export_commodity_codes ?

exportplan/management/commands/report_export_plan.py
```diff
@@ -0,0 +1,72 @@
...
+            # With country or commodity checks.
+            if plan.export_commodity_codes or plan.export_countries:
+                self.stdout.write(
+                    self.style.SUCCESS(f'Picked Product: {plan.export_commodity_codes[0]["commodity_name"]}')
+                ) if plan.export_commodity_codes else None
+                self.stdout.write(
+                    self.style.SUCCESS(f'Picked Country: {plan.export_countries[0]["country_name"]}')
+                ) if plan.export_countries else None
+
+                if is_ep_plan_empty(plan, set_useable_fields()):
```
depsiatwal, line 35: is_ep_plan_empty(plan, set_useable_fields()):
seems to be called in the if and also in the else. Hence it's called every time hence why not have the check outside the if statements as that's what is effectively happening here

depsiatwal (commented):
tests and check comments please

exportplan/management/commands/report_export_plan.py
seijihg: updated this part so it will be called only once.

exportplan/management/commands/report_export_plan.py
seijihg: its commodity code  or country so if one is true it will hit that code. If none is filled it will skip this part. In the update code this part is just for stdout really.

seijihg:
> tests and check comments please

Updated and ready to test! :D hope its good now.

exportplan/management/commands/report_export_plan.py
```diff
@@ -0,0 +1,69 @@
...
+        export_plans = models.CompanyExportPlan.objects.all()
+
+        for plan in export_plans.iterator():
+
+            self.stdout.write(self.style.SUCCESS(f'{plan.company}:')) if plan.company else self.stdout.write(
+                self.style.WARNING("No Company is associated with this EP.")
+            )
+
+            # Terminal stdout for Picked Product or Country.
+            if plan.export_commodity_codes or plan.export_countries:
```
depsiatwal, line 27: this is likely to fail 
plan.export_commodity_codes[0]
if we have countries due to the check.

exportplan/management/commands/report_export_plan.py
```diff
@@ -0,0 +1,69 @@
...
+
+        product_country_no_data = 0
+        product_country_with_data = 0
+
+        export_plans = models.CompanyExportPlan.objects.all()
+
+        for plan in export_plans.iterator():
+
+            self.stdout.write(self.style.SUCCESS(f'{plan.company}:')) if plan.company else self.stdout.write(
+                self.style.WARNING("No Company is associated with this EP.")
```
depsiatwal, line 23: ahh I see this is an conditional check

depsiatwal (approved):
LGTM

PR #931, opened 2021-08-18
GP2-2894 Migrate existing costs and price data

froddd:
Related to: https://github.com/uktrade/great-cms/pull/1492)

This migration will update the `total_costs_and_price` JSON blob for existing export plans to follow the new format:
- `units_to_export_first_period` is now `export_quantity`
- `units_to_export_second_period` is now `export_end` and formatted as an object with `month` and `year` property. Those are derived form the export plan record modified date and a relative delta based on the original `value` and `unit` for `units_to_export_second_period`.

Existing keys are left in place for reference, should this data need to be revised again in the future.

 - [ ] Changelog entry added.


depsiatwal (approved):
LGTM

PR #935, opened 2021-09-03
gp2-3351-auth-owner-only

depsiatwal:
Permissions class that allows only the owner to edit or change an exportplan

 - [x] Changelog entry added.

### DIFF
--- a/exportplan/permissions.py
+++ b/exportplan/permissions.py
@@ -9,10 +9,6 @@ class IsExportPlanOwner(permissions.BasePermission):
     """
 
     def has_permission(self, request, view):
-        try:
-            get_id = view.kwargs.get('pk') or request.data.get('companyexportplan')
-            export_plan = CompanyExportPlan.objects.get(pk=get_id)
-            return export_plan.sso_id == request.user.id
-        except CompanyExportPlan.DoesNotExist:
-            # Allow this caller should fail since no export plan with this ID exists
-            return True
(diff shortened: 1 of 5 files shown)


exportplan/permissions.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+
+    def has_permission(self, request, view):
+        try:
+            get_id = view.kwargs.get('pk') or request.data.get('companyexportplan')
+            export_plan = CompanyExportPlan.objects.get(pk=get_id)
+            return export_plan.sso_id == request.user.id
+        except CompanyExportPlan.DoesNotExist:
+            # Allow this caller should fail since no export plan with this ID exists
+            return True
```
bobmeredith, line 18: Er shouldn't this be False.
True, the user can't access an export plan if it doesn't exist with this id, but whatever happens, we should make the notfound and sso_id not match errors exactly the same, otherwise it potentially allows a way for users to fish for valid ids.

depsiatwal: i'm seeing this class as very atomic. If we return false that would subsequently have a start us 403 which is incorrect as opposed to 404. Therefore if we allow restapi to continue it will return a 404 to the caller. 

Of course there is one rare race condition where the ID may be created and the call returns to the restapi and bad user is able to update the export plan. We could filter on the query and use sso_id which would prevent this and return 404.

bobmeredith (approved):
LGTM

PR #938, opened 2021-09-11
Gp2 3179 - Spike additional checks added to the code.

seijihg:
Adds additional checks/tests to make sure code wont fail during stdout.

 - [x] Changelog entry added.


depsiatwal (approved):
LGTM

PR #939, opened 2021-09-13
Gp2 3179

seijihg:
Reports written into CSV:
Screenshot 2021-09-13 at 22 53 18
Screenshot 2021-09-13 at 22 53 23

 - [x] Changelog entry added.


depsiatwal (commented):
don't think you need this can hopefully use the pipe command

make manage MYCOMMAND >> MYFILE.txt

PR #942, opened 2021-09-16
Unsubscribe via token

rafa-garcia:
This PR changes the way users unsubscribe via hyperlink, introducing the parameters `uidb64` and `token` for anonymity in the query string. 

To do:

 - [X] Changelog entry added.


depsiatwal (approved):
LGTM

PR #943, opened 2021-09-16
Make unsubscribe backwards compatible

rafa-garcia:
To do:

 - [X] Changelog entry added.


depsiatwal (approved):
any changes in client ?

rafa-garcia:
@depsiatwal the client does not need to change as it is version-based. I think it is good that we keep an explicit parameter allowance on the client.

PR #944, opened 2021-09-16
GP2-3358-Fix-for-Dev

seijihg:
Fix to loop to make sure when both Country and Product exist it will register and not count towards Country OR Product.

Screenshot 2021-09-16 at 23 05 25

 - [x] Changelog entry added.


depsiatwal (approved):
lg

PR #945, opened 2021-09-16
GP2-3358-Fix-for-Prod

seijihg:
Fix to loop to make sure when both Country and Product exist it will register and not count towards Country OR Product.

 - [x] Changelog entry added.


depsiatwal (approved):
lgtm

appears one of the lines are too long please check

depsiatwal (approved):
remember this will need to be ported to prod

PR #951, opened 2021-09-30
GP2-3405 Export plan name with default number

bobmeredith:
Adds a name field to the export plan model.
On creating a new EP, the name is populated using the product and market in the EP data. The users' other plans are then checked for a clashing name and if there is a clash, a number in braces is added.
All matching EPs are checked so if an EP with a number is deleted, the next created one will slot in to the empty lowest number slot.

 - [x] Changelog entry added.


depsiatwal (approved):
LGTM

PR #952, opened 2021-09-30
Added S3 bucket for email enrolment images

webbyfox:

 - [x] Changelog entry added.


conf/settings.py
```diff
@@ -312,6 +312,11 @@
 AWS_STORAGE_BUCKET_NAME_DATA_SCIENCE = env.str('AWS_STORAGE_BUCKET_NAME_DATA_SCIENCE', '')
 AWS_S3_REGION_NAME_DATA_SCIENCE = env.str('AWS_S3_REGION_NAME_DATA_SCIENCE', '')
 
+# Setting for email buckets which holds images
+AWS_ACCESS_KEY_ID_EMAIL = env.str('AWS_ACCESS_KEY_ID_EMAIL', '')
+AWS_SECRET_ACCESS_KEY_EMAIL = env.str('AWS_SECRET_ACCESS_KEY_EMAIL', '')
+AWS_STORAGE_BUCKET_NAME_EMAIL = env.str('AWS_STORAGE_BUCKET_NAME_EMAIL', '')
```
depsiatwal, line 318: Why do we need another bucker, we now have 3 in total :-) 
I thinking we should make do with 1

depsiatwal (commented):
see inline comment

conf/settings.py
webbyfox: Good point, Don't think we can merge into one - as we have data science bucket. we already had three buckets - as directory-email-static bucket was hardcoded in constant.py, so using this opportunity to move in to vault. This change we are doing to move that bucket from Ireland(eu-west-1) to London (eu-west-2) but let me check with Jayesh if they have same permission level then we can move those email images to existing bucket

webbyfox: checked with Jayesh - That EMAIL bucket is only one bucket for all environments and those existing buckets have separate bucket for each environment

depsiatwal (approved):
As discussed all Good 
bucket is using an already existing bucket this is just moving it to config

PR #975, opened 2021-12-13
GP2-3891-itr1-push-ep-data-data-workspace

depsiatwal:

 - [ ] Changelog entry added.


rafa-garcia (approved):
LGTM

activitystream/views.py
```diff
@@ -261,3 +262,22 @@ def list(self, request):
...
+    """View set to list export plan for the activity stream"""
+
+    @decorator_from_middleware(ActivityStreamHawkResponseMiddleware)
+    def list(self, request):
+        """A single page of companies to be consumed by activity stream."""
+        after_ts, after_id = self._parse_after(request)
+
+        export_plans = list(
+            CompanyExportPlan.objects.filter(
+                Q(modified=after_ts, id__gt=after_id) | Q(modified__gt=after_ts),
```
bobmeredith, line 277: I'm not sure I'm reading this right. Q(a,b) is a and b so would only pick up new export plans if the modified date is exactly equal to after_ts and has id > after_id which seems very unlikely.
I guess I'm wondering how AS works out after_id as it's not the greatest id transferred or indeed the last one,

PR #1035, opened 2022-04-04
Bootstrap UK total trade by country endpoint

rafa-garcia:

 - [ ] Changelog entry added.


conf/urls.py
```diff
@@ -269,6 +269,11 @@
         dataservices.views.RetrieveSocietyDataByCountryView.as_view(),
         name='dataservices-society-data-by-country',
     ),
+    url(
+        r'^dataservices/uk-total-trade-data-by-country/(?P<iso2>.*)/$',
```
froddd, line 273: Nitpick: it seems all other endpoints take country as a query param rather than a named URL fragment. Is there a reason to deviate from that here?

PR #1041, opened 2022-04-27
Gls 186 highlights

rafa-garcia:
This pull request covers the handling of the [ONS' UK total trade dataset](https://www.ons.gov.uk/economy/nationalaccounts/balanceofpayments/datasets/uktotaltradeallcountriesseasonallyadjusted), ingested by quarterly activity, via two endpoints to yield market trends and trade highlights for a particular country.

To do:

 - [X] Changelog entry added.


dataservices/views.py
```diff
@@ -200,26 +199,64 @@ def get(self, *args, **kwargs):
...
-        return get_object_or_404(
-            models.Country,
-            iso2__iexact=iso2,
-        )
-
     def get(self, *args, **kwargs):
         res = super().get(*args, **kwargs)
-        country = self.get_country(self.request.query_params.get('iso2', ''))
+        if not res.data:
+            raise Http404
```
froddd, line 218: Not sure it's helpful to return a 404 here -- the resource does exist, but there is not data, so the API should probably return a 200 with something like:

```json5
{
  "metadata": {
    // metadata for the query here
  },
  "data": null
}
```

dataservices/views.py
```diff
@@ -200,26 +199,64 @@ def get(self, *args, **kwargs):
...
+        return obj
+
+    def get(self, *args, **kwargs):
+        iso2 = self.request.query_params.get('iso2', '')
+        if not iso2:
+            return Response(status=status.HTTP_400_BAD_REQUEST, data={"iso2": ['This field is required.']})
+
+        obj = self.get_object()
+        if not obj:
+            raise Http404
```
froddd, line 258: Same as above, a 404 seems odd here

dataservices/views.py
rafa-garcia: It's the `200` what bugs me when in the case an `iso2` code does not exist. The resource is a collection of trade records for a particular country, so when the country is not found in my mind a `404` seems more reasonable than a `200` with an empty data list.

froddd: The resource is `/dataservices/uk-market-trends/`, so trying to access `/dataservices/uk-market-trends/?iso2=VA` means:

- Resource exists -- if it didn't: would return `404`
- Country iso2 is valid -- if it wasn't present, or wasn't valid: invalid request, would return `400`
- Data does not exist either because country not in DB or data not available for this period or those parameters, return `200` with `null` data as the query can be run but there is not data returned by it

PR #1042, opened 2022-04-28
Update UK total trade datasets

rafa-garcia:
To do:

 - [X] Changelog entry added.


dataservices/management/commands/tests/test_import_data.py
```diff
@@ -222,4 +222,4 @@ def test_import_uk_total_trade_data():
     management.call_command('import_countries')
     management.call_command('import_uk_total_trade_data')
 
-    assert len(models.UKTotalTradeByCountry.objects.all()) == 4439
+    assert len(models.UKTotalTradeByCountry.objects.all()) == 4632
```
froddd, line 225: Is it useful to test the number of actual objects imported, versus simply testing the import does not throw? The other import tests seem to consider that sufficient, which means no need to update when a new dataset is added?

rafa-garcia: To be honest, this kind of tests are time consuming and bound to the content of the file source which changes quite a bit. But I'm afraid we need to ensure a certain level of integrity, because we are not ingesting all of the records in the file and columns are being transposed to rows.

A better approach in my opinion would be to mock the binary file and yield a few rows that cover different cases, quick and steady.

froddd: Yeah, a reproducible test case would be better. If there are several filters to apply to the source file, then a test suite with separate tests for each of them (abstracting the filtering to a separate, unit-testable method) would be ideal, though possibly overkill. A test input file should be able to cover all cases at once, as long as the filters are properly documented somewhere (would someone picking up this codebase understand each filter/exclusion?).

PR #1071, opened 2022-08-02
Add IMF's World Economic Outlook to dataservices

rafa-garcia:
This pull request covers the handling of the [IMF's  World Economic Outlook (WEO) database](https://www.imf.org/en/Publications/SPROLLs/world-economic-outlook-databases) from ingestion via the [Data Workspace's SQL API](https://data.trade.gov.uk/datasets/14511382-b622-44cf-a249-259c04e0b6c7), to the implementation of an API endpoint which yields economic highlights for a particular country.

An example `GET` request to `/dataservices/economic-highlights?iso2=cn` would return the latest economic outlook for China (non-projected) in terms of GDP per capita (U.S. Dollars) and economic growth (percentage) with raw values:

```JSON
{
    "metadata": {
        "country": {
            "name": "China",
            "iso2": "CN"
        },
        "uk_data": {
            "gdp_per_capita": {
                "year": 2020,
                "value": 41127.443,
                "is_projection": true
            }
        }
    },
    "data": {
        "gdp_per_capita": {
            "value": 10525.001,
            "year": 2020
        },
        "economic_growth": {
            "value": 8.08,
            "year": 2021
        }
    }
}
```

To do:

 - [X] Changelog entry added.


dataservices/views.py
```diff
@@ -311,3 +317,33 @@ def dispatch(self, *args, **kwargs):
...
+    permission_classes = []
+    queryset = models.WorldEconomicOutlookByCountry.objects
+    serializer_class = serializers.EconomicHighlightsSerializer
+    filter_class = filters.EconomicHighlightsFilter
+    lookup_field = 'country__iso2'
+
+    def get_metadata(self, **kwargs):
+        country = self.get_country()
+        year = kwargs['gdp_per_capita']['year']
+        uk_data = {'gdp_per_capita': self.queryset.uk_stats(self.queryset.GDP_PER_CAPITA_USD_CODE, year)}
```
mforner13, line 332: Should this have the UK economic growth stat too?

dataservices/views.py
```diff
@@ -215,11 +219,13 @@ def get_metadata(self):
 
     def get(self, *args, **kwargs):
         res = super().get(*args, **kwargs)
-        metadata = self.get_metadata()
         data = res.data
 
         if isinstance(data, list):
             data = data[: self.limit]
+            metadata = self.get_metadata()
```
mforner13, line 226: I've probably missed something, but what will happen at this line from the new view class (line 331) work if `data` isn't a list?
`year = kwargs['gdp_per_capita']['year']`

mforner13 (commented):
Just two comments! I was also curious - is the plan for us to fully replace the data in `WorldEconomicOutlookByCountryFactory` using the management command (I've seen it deletes all objects in the table at the end), rather than updating the objects?

dataservices/views.py
rafa-garcia: This endpoint won't be used to compare the economic growth of a country against UK's

mforner13: Ahh! Why is that? Is the data for that coming from elsewhere?

dataservices/views.py
rafa-garcia: `data` is only needed when the object is NOT a list. In this instance, we need to have the `data` first so we have the latest available year of GDP, in order to query for UK's stats for this particular metric and year.

dataservices/views.py
rafa-garcia: `data` comes from calling `stats()` on the queryset. It returns, among other things, the latest year worth of GDP per capita of a particular country. We then use that `year` value to call `uk_stats(code, year)` on the same queryset to get a similar payload but this time for the UK.

rafa-garcia: We want to compare the same metrics (GDP per capita) AND the same year of a country against UK's

rafa-garcia: @mforner13 I misunderstood your question. The economic growth of a country won't have a UK reference as per https://www.great.gov.uk/markets/china/ (at a glance section) for instance

mforner13: Ah yes just noticed that! I'd assumed we show UK data for economic growth as well - all good 👍

mforner13 (approved):
LGTM 👍

rafa-garcia:
> Just two comments! I was also curious - is the plan for us to fully replace the data in `WorldEconomicOutlookByCountryFactory` using the management command (I've seen it deletes all objects in the table at the end), rather than updating the objects?

I understand that you mean `WorldEconomicOutlookByCountry` rather than the factory. We play safe here by deleting the dataset and re-populating the table, because we will have different number of records from time to time.

PR #1072, opened 2022-08-16
Add last release date to dataset as metadata

rafa-garcia:
Information about the last release date of certain datasets is required to be refreshed in an automated fashion, and be served as a key-value pair inside a `metadata` object in the responses of API calls to endpoints that use those datasets.

This pull request brings the means to ingest this data from Data Workspace, in addition to presenting a `Metadata` class to handle any metadata key-value pairs from the Django admin site.

image

image

image

To do:

 - [X] Changelog entry added.
 - [X] Requirements have been compiled.


tomstevens9 (approved):

tomstevens9:
LGTM 👍

PR #1074, opened 2022-08-26
Update package to use Django 3.2

tomstevens9:

 - [x] Changelog entry added.


mforner13 (approved):
Doing the lord's work 🙌

PR #1101, opened 2022-12-07
Add missing fields & extra text type to surveys

mforner13:
This PR adds a couple of thing which I missed in #1099, which are:

- Returning the question `type` and `add_other_option` fields in the survey serializer
- Adding more specific text options for question types - `long_text` and `short_text`

 - [ ] Changelog entry added.


rafa-garcia (approved):
👍🏻

PR #1112, opened 2023-01-24
Fix ViewSet attribute to invoke filtersets

rafa-garcia:
[django-filter](https://github.com/carltongibson/django-filter) renamed `ViewSet.filter_class` to `ViewSet.filterset_class` [some time ago](https://github.com/carltongibson/django-filter/pull/867). The deprecated attribute name remained valid until v22 but now results in filters not being applied.

To do:

 - [X] Changelog entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Pre release
 ### Bugs fixed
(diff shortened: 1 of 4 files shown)


ecbm (approved):
👍

PR #1120, opened 2023-02-17
Add logic for multi-select additional routing

mforner13:
This PR adds validation to both `ChoiceInlineAdmin` and `ChoiceAdmin` to prevent multiple choices being added to a multi-select question that have additional routing.

The validation is needed in both classes, as the logic is different when changing multiple Choices at once (the inline formset) and one Choice (the regular admin form).

Validation has not been added to the model, as this causes errors when saving multiple choices at once via the inline form, as the choice being saved, is not aware of updates to the other choice objects, and so cannot determine whether another choice on the question has additional routing or not.

As surveys will only be added via the django admin interface, I think this is an acceptable choice.

*NOTE*
The PR also includes a missing migration - which updates the line length in some choice field help texts - to make flake8 happy.

**Inline admin formset**
Screenshot 2023-02-17 at 16 35 26

**Admin form**
Screenshot 2023-02-17 at 16 33 46

 - [x] Changelog entry added.


rafa-garcia (approved):
LVGTM 👍🏻

PR #1126, opened 2023-03-08
Allow multiple region_ids for one Office object

mforner13:
This PR adds a new field `region_ids` to the Office model. This is due to the consolidation of regional offices, meaning that multiple region ids will relate to one Trade office. NOTE - region IDs come from the postcode API.

The next steps in this change are:
- Once released, move the data in `region_id` to `region_ids`
- Make sure all `region_id` fields are empty
- Create a new migration to remove the `region_id` field and to make the new `region_ids` field mandatory.
- Update the logic in the view to only check the region_ids field.

**The new field in django admin**
Screenshot 2023-03-08 at 15 33 41

 - [x] Changelog entry added.


exporting/models.py
```diff
@@ -8,11 +9,12 @@ class Meta:
...
-    region_id = models.TextField()
+    region_id = models.TextField(blank=True, null=True)
+    region_ids = ArrayField(models.TextField(), null=True, help_text='Regions need to be separated by commas.')
     name = models.TextField()
-    address_street = models.TextField()
-    address_city = models.TextField()
-    address_postcode = models.TextField()
+    address_street = models.TextField(blank=True, null=True)
+    address_city = models.TextField(blank=True, null=True)
+    address_postcode = models.TextField(blank=True, null=True)
```
timothyPatterson, line 17: Perhaps a max_length property of 8 (or 7 and strip out whitespace) to provide some validation? This is because the  PostCode is pivotal to the searching via helpers.postcode_to_region_id method.

timothyPatterson (commented):
Tested locally and working well. Small question about validating postcode given it is critical to how this works.

PR #1184, opened 2023-08-31
UAT and production release

TOE703:


conf/settings.py
```diff
@@ -287,6 +287,7 @@
 COLLABORATOR_INVITE_SUBJECT = env.str(
     'COLLABORATOR_INVITE_SUBJECT', 'Confirm you’ve been added to {company_name}’s Find a buyer profile'
 )
+GREAT_MARKETGUIDES_TEAMS_CHANNEL_EMAIL = env.str('GREAT_MARKETGUIDES_TEAMS_CHANNEL_EMAIL')
```
timothyPatterson, line 290: Would a default here be useful, e.g. 'test@example.com' as per dev conf - mainly for consistency with the other env. variables ?

timothyPatterson (approved):

PR #1186, opened 2023-09-07
Add an endpoint for Activity Stream to consume export plan data

rafa-garcia:
CONTEXT: This PR introduces an API endpoint that exposes export plan data for Activity Stream to consume.

The data structure has been designed to match the sections with all of the question groups available in the [web forms](https://www.great.gov.uk/export-plan/).

The model includes an `answers_count` dynamic property which mirrors the logic of the progress indicator on the frontend. This property is reflected in the JSON payload, making it easy to filter by completion degree of export plans.

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/KLS-1018
- [x] Jira ticket has the correct status
- [x] A clear/description pull request messaged added

### Merging

- [x] This PR can be merged by reviewers


activitystream/filters.py
```diff
@@ -0,0 +1,12 @@
+import datetime
+
+from django_filters import CharFilter, FilterSet
+
+
+class ActivityStreamCompanyExportPlanFilter(FilterSet):
+    after = CharFilter(method='filter_after')
```
sebs-code, line 7: Might be good to have a doc string on these (high level) explaining what they do.

rafa-garcia: I agree that it's a good practice, however, we're dealing with the simplest, standard `django-filter` filtersets which is wholly documented.

activitystream/views.py
```diff
@@ -316,25 +320,20 @@ def list(self, request):
         )
 
 
-class ActivityStreamExportPlanDataViewSet(BaseActivityStreamViewSet):
-    """View set to list export plan data for the activity stream"""
+class ActivityStreamExportPlanView(ListAPIView):
+    authentication_classes = (ActivityStreamAuthentication,)
+    permission_classes = ()
```
sebs-code, line 325: Do we need an empty permission_classes tuple here? Won't it be open by default?

activitystream/filters.py
rafa-garcia: Do you mean the `filter_after` method, though? I pulled it from an AS view and that's undocumented. I'll put up a comment

activitystream/views.py
rafa-garcia: I'm putting back `permission_classes` with an empty tuple because otherwise the responses are `403`

sebs-code (approved):
Thanks for response to queries. Looks good.

PR #1329, opened 2024-11-04
Feature greatuk 1286 add get request caching

bobby-didcoding:
This PR is to implement GET request caching in directory-api and cache invalidation into POST, and PUT requests.

Testing:

TBC

### Workflow

- [x] Ticket exists in Jira https://uktrade.atlassian.net/browse/GREATUK-1286
- [x] Jira ticket has the correct status.
- [x] A clear/description pull request messaged added.

### Merging

- [x] This PR can be merged by reviewers.


conf/settings.py
```diff
@@ -82,6 +84,7 @@
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'django.middleware.cache.FetchFromCacheMiddleware',
```
timothyPatterson, line 87: It seems from docs that this always needs to be last in the middleware list. It might be helpful to add a comment to this effect in case future middleware is added beneath. [source](https://docs.djangoproject.com/en/5.1/topics/cache/#order-of-middleware)

timothyPatterson (approved):

conf/settings.py
bobby-didcoding: Good point! I've just added some new code.

